PR #28, opened 2019-06-05
Code coverage in CI loop

bjk17:
Changes:
 - calculating code coverage in CI loop and pushing to Coveralls.io (not sure how the coverage have gotten there until now)
 - moved test dependencies to the `setup.py` file and deleted the corresponding `.txt` file
 - some formatting and typo in a Python file I stumbled upon

Notes:
 - should pinpoint the versions of the packages in `install_requires` in `setup.py` to ensure consistency between installations of the `Tilings` package


enadeau:
Changed it to merge into develop We should only merge into master for a new release.

Regarding the version specification, should we just point to the most recent version on pypi for now?

bjk17:
> Looks good. Can you also add the badge on the readme?

It's there already.

bjk17:
> Changed it to merge into develop We should only merge into master for a new release.

Thank you! My bad.

> Regarding the version specification, should we just point to the most recent version on pypi for now?

Yes, that would be a good starting point. Later we can start using a service called [requires.io](https://requires.io) that keeps track of dependencies getting too old (with a badge in README).

PR #32, opened 2019-08-11
Add method is_empty_cell and is_monotone cell

enadeau:


enadeau:
@christianbean Does the `is_empty_cell` check needed in the `is_monotone_cell` method or the presence of a longer local obstruction guaranty that there's no `1` obstruction in the cell?

enadeau:
@christianbean volunteer to fix the requirements part of the `cell_basis` method

PR #41, opened 2019-08-28
Fusion

enadeau:
Here is the new fusion on Tilings!


enadeau:
I'll try to run some real life test to compare with the old tilescope fusion algorithm

enadeau:
I made a run of tilescope when every time it goes in the `fusion` function it compares the output of the new algorithms and the ones that currently in tilescope. I got the tree for 123, 1234, and 12345 and the algorithm didn't any tiling where the output were different. I also ran 1324 and 1432_2143 for a while without errors.

enadeau:
Good to go!

PR #43, opened 2019-08-30
optimize containment test for gridded perm

enadeau:
This is ready to go. Before merging, we just need to wait that permuta release the needed feature


ulfarsson:
So we need to release a new version of Permuta to PyPI for the tests to pass?

christianbean:
Emile had a fix for this in Tilescope pulling the correct branch of Tilings. This can be done for Tilings and Permuta

PR #44, opened 2019-09-06
Factor

enadeau:


enadeau:
This is mostly ready to merge, we just need to see if we need to implement union factor or if it will be covered by the new feature we are adding to tilescope.

enadeau:
I think this one is ready for review

tilings/algorithms/factor.py
```diff
@@ -0,0 +1,280 @@
...
+                for f in self._get_factors_obs_and_reqs()]
+
+    def reducible_factorisations(self):
+        """
+        Iterator over all reducible factorisation that can be obtained by
+        grouping of irreducible factors.
+
+        Each factorisation is a list of tiling.
+
+        For example if T = T1 x T2 x T3 then (T1 x T3) x T2 is a possible non
```
ulfarsson: non-irreducible -> reducible

tilings/algorithms/factor.py
```diff
@@ -0,0 +1,280 @@
...
+        irreducible factorisation.
+        """
+        min_comp = self._get_factors_obs_and_reqs()
+        print(min_comp)
```
ulfarsson: No printing please

tilings/algorithms/factor.py
```diff
@@ -0,0 +1,280 @@
...
+        Returns the type of constructor for the factorisation
+        """
+        return 'cartesian'
+
+    def _rule(self, factors, formal_step, workable):
+        """
+        Return the comb_spec_searcher rule for the factorisation where the list
+        of factors is given as a list of tilings.
+
+        TODO: Describe the meaning or workable.
```
ulfarsson: If workable=True, then we expand the children, and want to ignore the parent. If workable=False, then we do not expand the children, and want to keep working on the parent (perhaps placing points into cells or something).

tilings/algorithms/factor.py
```diff
@@ -0,0 +1,280 @@
...
+                    inferable=[False for _ in factors],
+                    workable=[workable for _ in factors],
+                    possibly_empty=[False for _ in factors],
+                    ignore_parent=workable,
+                    constructor=self.constructor)
+
+    def rule(self, workable=True):
+        return self._rule(self.factors(), self.formal_step(), workable)
+
+    def all_union_rules(self, workable=True):
```
ulfarsson: By default workable=False in all unions. But work from the maximal factorization.

tilings/algorithms/factor.py
```diff
@@ -0,0 +1,280 @@
...
+    Two non-empty cells are in the same factor if they share an obstruction or
+    a requirement.
+    """
+
+    def _unite_rows_and_cols(self):
+        """
+        Override the `Factor._unite_rows_and_cols` to do nothing since
+        interleaving is allowed on row and column.
+        """
+        pass
```
ulfarsson: Perhaps instead overwrite the _unite_all function?

enadeau: I think we should let this one this way since the behaviour of interleaving factor is to not union any row and columns. If don't do anything the behaviour from the parent class will apply. I'll will also overide the _unite_all method.

Issue #52, opened 2019-09-13
Obstruction inferral


enadeau:
- [x] Make the algorithm
  - [x] Subobstruction Inferral
  - [x] Empty cell inferral
  - [x] All obstruction inferral
- [x] Make the method on tilings
  - [x] All obstruction inferral
  - [x] subobstruction inferral
  - [x] empty cell inferral

No strategy will be made for the moment.

PR #61, opened 2019-09-13
Obstruction inferral

enadeau:
Address issue #52 

I'm still working on all obstruction inferral and the tiling methods.

### DIFF
--- a/tilings/algorithms/obstruction_inferral.py
+++ b/tilings/algorithms/obstruction_inferral.py
@@ -17,14 +17,14 @@ class ObstructionInferral(abc.ABC):
     @abc.abstractmethod
     def potential_new_obs(self):
         """
-        Return an iterable of new obstruction that should be added to the
-        tiling if possible
+        Return an iterable of new obstructions that should be added to the
+        tiling if possible.
         """
         pass
 
     def new_obs(self):
         """
-        Returns the list of new obstruction that can be added to the tiling.
+        Returns the list of new obstructions that can be added to the tiling.
         """
         if hasattr(self, '_new_obs'):
             return self._new_obs
@@ -38,7 +38,7 @@ class ObstructionInferral(abc.ABC):
 
     @staticmethod
     def can_add_obstruction(obstruction, tiling):
-        """Return true if obstruction can be added to tiling."""
+        """Return true if `obstruction` can be added to `tiling`."""
         return (tiling.add_requirement(obstruction.patt, obstruction.pos)
                 .is_empty())
 
@@ -77,7 +77,7 @@ class SubobstructionInferral(ObstructionInferral):
         """
         subobs = set()
         for ob in self._tiling.obstructions:
-            subobs.update(ob.all_subperms())
+            subobs.update(ob.all_subperms(proper=True))
         subobs.remove(Obstruction(Perm(), []))
         return subobs
 
--- a/tilings/algorithms/obstruction_inferral.py
+++ b/tilings/algorithms/obstruction_inferral.py
@@ -114,16 +114,11 @@ class AllObstructionInferral(ObstructionInferral):
         """
         Iterator over all possible obstruction of `self.obstruction_length`.
         """
-        active = self._tiling.active_cells
+        no_req_tiling = self._tiling.__class__(self._tiling.obstructions)
         n = self._obs_len
-        all_gp = (Obstruction(patt, pos) for patt, pos in
-                  product(PermSet(n), product(active, repeat=n)))
-        pot_obs = all_gp
-        pot_obs = (gp for gp in all_gp if
-                   not gp.contradictory() and
-                   self.avoids_obstructions(gp) and
-                   self.not_required(gp))
-        return pot_obs
+        pot_obs = filter(self.not_required,
+                         no_req_tiling.gridded_perms_of_length(n))
+        return (Obstruction(gp.patt, gp.pos) for gp in pot_obs)
 
 
 class EmptyCellInferral(AllObstructionInferral):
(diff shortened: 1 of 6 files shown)


enadeau:
I'm done with the work on this. However, the implementation for all obstruction inferral is implemented very naively at the moment. Any suggestion to improve it is welcome. I might have one that we can discuss but I'm not sure it is worth our time at the moment.

tilings/tiling.py
```diff
@@ -724,6 +726,31 @@ def obstruction_transitivity(self):
         obs_trans = ObstructionTransitivity(self)
         return obs_trans.obstruction_transitivity()
 
+    def all_obstruction_inferral(self, obstruction_lenght):
```
christianbean: lenght -> length

tilings/algorithms/obstruction_inferral.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def not_required(self, gp):
+        """
+        Returns True if the gridded perm `gp` is not required by any
+        requirement list of the tiling.
+        """
+        return all(any(gp not in req for req in req_list)
+                   for req_list in self._tiling.requirements)
+
+    def potential_new_obs(self):
```
christianbean: I would suggest doing 
```
yield from Tiling(obstructions).gridded_perms_of_length(length)
```
It should achieve the same thing with far less explosion.

tilings/algorithms/obstruction_inferral.py
```diff
@@ -0,0 +1,149 @@
...
+                    requirements=merged_tiling.requirements,
+                    remove_empty=False)
+        self._new_obs = newobs
+        return self._new_obs
+
+    @staticmethod
+    def can_add_obstruction(obstruction, tiling):
+        """Return true if obstruction can be added to tiling."""
+        return (tiling.add_requirement(obstruction.patt, obstruction.pos)
+                .merge().is_empty())
```
christianbean: The merge isn't necessary. I would remove that.

tilings/tiling.py
enadeau: Done

tilings/algorithms/obstruction_inferral.py
```diff
@@ -0,0 +1,149 @@
...
+            return self._new_obs
+        newobs = []
+        merged_tiling = self._tiling.merge()
+        for ob in sorted(self.potential_new_obs(), key=len):
+            if self.can_add_obstruction(ob, merged_tiling):
+                newobs.append(ob)
+                merged_tiling = merged_tiling.__class__(
+                    obstructions=chain(merged_tiling.obstructions, (ob,)),
+                    requirements=merged_tiling.requirements,
+                    remove_empty=False)
```
christianbean: Rather than creating a new tiling, it is probably faster to just check if it avoids all of the newobs. In fact, you probably want to do between line 33 and 34.

tilings/algorithms/obstruction_inferral.py
```diff
@@ -0,0 +1,149 @@
...
+        pass
+
+    def new_obs(self):
+        """
+        Returns the list of new obstruction that can be added to the tiling.
+        """
+        if hasattr(self, '_new_obs'):
+            return self._new_obs
+        newobs = []
+        merged_tiling = self._tiling.merge()
```
christianbean: Avoid merging. See comment on line 47. The is empty function is deterministic. Merging relies on the generation of gridded perms exhaustively, whereas relying on just is empty is a fail-fast approach.

tilings/algorithms/obstruction_inferral.py
enadeau: I think that this suggestion doesn't work. Take the following code
```python
t = Tiling(obstructions=[
    Obstruction(Perm((0, 1, 2)), ((0, 0),)*3),
    Obstruction(Perm((0, 1, 2)), ((1, 0),)*3),
    Obstruction(Perm((0, 1)), ((0, 0), (1, 0))),
    Obstruction(Perm((1, 0)), ((0, 0), (1, 0))),
], requirements=[[
    Requirement(Perm((0,)), ((1, 0),)),
]])
print(t)
print('Contained gridded permutations of length 1')
for p in t.gridded_perms_of_length(1):
    print(p)
```
The output is
```
+-+-+
|1|2|
+-+-+
1: Av(012)
2: Av+(012)
Crossing obstructions:
01: (0, 0), (1, 0)
10: (0, 0), (1, 0)
Requirement 0:
0: (1, 0)
Contained gridded permutations of length 1
0: (1, 0)
```
With your suggestion we would never try to add the obstruction `0: (0, 0)`

tilings/algorithms/obstruction_inferral.py
enadeau: I agree. I wasn't sure if it was needed.

tilings/algorithms/obstruction_inferral.py
christianbean: We would, note, I am suggesting initialising without the requirements.

christianbean: A thought: 

Let T = (O, R) be a tiling, and T' = (O, emptyset).

Is Grid(T') - Grid(T) precisely those we want to add?

tilings/algorithms/obstruction_inferral.py
enadeau: Did it

tilings/algorithms/obstruction_inferral.py
enadeau: Changed it

tilings/algorithms/obstruction_inferral.py
christianbean: This is obviously wrong... (e.g every one by one positive tiling becomes empty...). 

However, all obstructions we can add do satisfy o in Grid(T') and o not in Grid(T).

enadeau: I pushed a more clever filtering that includes what we've discussed

PR #65, opened 2019-09-26
Verification strategy

### LINKED ISSUE #55: Globally, subset and point verify
The algorithm
- [x] `GloballyVerified` class
- [x] `SubsetVerified` class

The method
- [ ] `Tiling.subset_verified`
- [ ] `Tiling.globally_verified`

The strategy
- [ ] `subset_verified`
- [ ] `globally_verified`
- [ ] `point_verified`

We also want a way to get the enumeration of the verified tilings and possibly a tree that described them.

enadeau:
Closes #55


enadeau:
For the moment we can't merge this pull request since the dependencies on tilescope to find tree for verified objects creates circular dependencies.

enadeau:
For the moment we can't merge this pull request since the dependencies on tilescope to find tree for verified objects creates circular dependencies.

enadeau:
I'm mostly done with this pull request. The only missing enumeration is `OneByOneEnumeration`. The part about finding tree should be done later in another PR see #68. 

I think I'll need of help of @christianbean to update the `get_genf` method of tiling.

Also, we should decide if we want to be able to check if a tiling is verified with a method. I see three option:
- a method for each verification strategy (e.g.. `Tiling.locally_enumerable`, `Tiling.locally_factorable`, ...)
- a single method `Tiling.verified` that checks if any of the verification strategy applies
- no verification method on Tiling

enadeau:
We decide to have a single method that returns the verification strategy that apply.

PR #79, opened 2019-12-13
[requires.io] dependency update on develop branch

enadeau:


enadeau:
The version of sympy can't be updated until we update the version of comb_spec_searcher.

PR #83, opened 2020-01-31
Move the strategies to tilings

enadeau:
Closes #83 

- [x] Verification strategies
- [x] Inferral strategies
- [x] Decomposition strategies
- [x] Equivalence strategies
- [x] Batch strategies
- [ ] Move the tests for the strategies' function from `tilescope` to `tilings`

### DIFF
--- a/tilings/strategies/batch.py
+++ b/tilings/strategies/batch.py
@@ -123,6 +123,17 @@ def col_placements(tiling, **kwargs):
     yield from RequirementPlacement(tiling).all_col_placement_rules()
 
 
+def all_placements(tiling, **kwargs):
+    req_placements = (RequirementPlacement(tiling),
+                      RequirementPlacement(tiling, own_row=False),
+                      RequirementPlacement(tiling, own_col=False))
+    for req_placement in req_placements:
+        yield from req_placement.all_point_placement_rules()
+        yield from req_placement.all_requirement_placement_rules()
+        yield from req_placement.all_col_placement_rules()
+        yield from req_placement.all_row_placement_rules()
+
+
 def row_and_col_placements(tiling, **kwargs):
     req_placements = RequirementPlacement(tiling)
     yield from req_placements.all_row_placement_rules()
--- a/tilings/strategies/batch.py
+++ b/tilings/strategies/batch.py
@@ -84,7 +84,7 @@ def all_requirement_insertions(tiling: Tiling, maxreqlen: int = 1,
                                  extra_basis).rules(ignore_parent)
 
 
-def all_factor_insertions(tiling: Tiling, ignore_parent: bool = False,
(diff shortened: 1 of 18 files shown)


enadeau:
-Is there any situation where we want to ignore parent of a requirement insertion strategy? I've ported it but I'm not sure we need it

tilings/strategies/equivalence.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    Strategy that place a single forced point of a point requirement.
+
+    The point placement strategy considers all point requirements in their own
+    requirement lists. For each of them, it returns a new tiling where the
+    point has been placed with a force.
+    """
+    yield from partial_requirement_placement(tiling, point_only=True)
+
+
```
christianbean: Fusion strategies should have a seperate file as they are not really "equivalence" strategies.

tilings/strategies/batch.py
```diff
@@ -0,0 +1,163 @@
...
+def partial_row_and_col_placements(tiling, **kwargs):
+    req_placements = (RequirementPlacement(tiling, own_row=False),
+                      RequirementPlacement(tiling, own_col=False))
+    for req_placement in req_placements:
+        yield from req_placement.all_row_placement_rules()
+        yield from req_placement.all_col_placement_rules()
+
+
+# TODO: Clarify what is this function doing
+def requirement_list_placement(tiling, **kwargs):
```
enadeau: I do not really understand what this strategy is doing? Can we clarify the docstring?

tilings/strategies/equivalence.py
enadeau: They were all in equivalence strategies in tilescope but I also prefer having its own file. I'll change that.

enadeau: Done.

tilings/strategies/batch.py
enadeau: We decided to remove it

tilings/strategies/equivalence.py
```diff
@@ -0,0 +1,68 @@
...
+def all_placements(tiling, **kwargs):
+    req_placements = (RequirementPlacement(tiling),
+                      RequirementPlacement(tiling, own_row=False),
+                      RequirementPlacement(tiling, own_col=False))
+    for req_placement in req_placements:
+        yield from req_placement.all_point_placement_rules()
+        yield from req_placement.all_requirement_placement_rules()
+        yield from req_placement.all_col_placement_rules()
+        yield from req_placement.all_row_placement_rules()
+
```
christianbean: Should all placements be in batch ? it has both batch and equivalence.

enadeau: Henning and I agreed with that. I changed it.

PR #89, opened 2020-02-21
Get genf

### LINKED ISSUE #68: Implement tree finding for verified tilings
When we port all the `tilescope` stuff to this we need to add pack to the enumerations algorithms.

enadeau:
This closes #68 

There's still some work to do with this PR, Especially regarding 1x1 tilings

### DIFF
--- a/tests/algorithms/test_enumeration.py
+++ b/tests/algorithms/test_enumeration.py
@@ -3,7 +3,7 @@ import abc
 import pytest
 import sympy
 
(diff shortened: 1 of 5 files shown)


enadeau:
Most of the work on this PR is done. We only miss the enumeration for the 1x1. This can be done until we get a new database with permutation class set up. As the rest of the code is useful on its own I suggest that we merge it and deal with the 1x1 later

PR #99, opened 2020-03-09
Tree expansion

enadeau:
This PR aims at adding the tool needed to perform tree expansion. The tools that needs to be added are:
- [x] a `MonotoneTreeVerificationStrategy`
- [ ] a method on `TileScopePack` to turn a pack into an expansion pack.


enadeau:
If we have monotone tree verification already merge from somewhere else we can close that

PR #100, opened 2020-03-10
Adding a finite verification strategy

enadeau:


enadeau:
We don't need it at the moment. we might merge this later

PR #105, opened 2020-03-15
Improving gridded perm generation

christianbean:
This will 

- [x] Make the `gridded_perms` function a class in the algorithms folder.

- [x] Keep track of the number of points a cell must require.

- [x] Add a minimal flag, that will only consider adding a point if it can lead to a longer requirement


enadeau:
I changed the name to `GriddedPermsOnTiling`

PR #106, opened 2020-03-19
Merge

christianbean:
- [x] a minimal gridded perm function

- [x] a new is empty function


tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -0,0 +1,123 @@
...
+            counter = defaultdict(int)
+        for cell in gp.pos:
+            counter[cell] += 1
+        return counter
+
+    def get_cells_to_try(
+                         self, gp: GriddedPerm, max_cell_counts: List[Counter],
+                         cells: List[Cell]) -> Iterator[Cell]:
+        """Yield cells that a gridded permutation could be extended by."""
+        currcellcounts = self.cell_counter(gp)
```
enadeau: For example if you used a `Counter` instead of `defaultdict` here you could just write `currcellcounts = Counter(gp.pos)`

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -0,0 +1,123 @@
...
+        if any(all(count >= max_cell_count[cell]
```
enadeau: Can you explain me why this `all` is not an `any`?

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -0,0 +1,123 @@
...
+            return
+
+        self.prepare_queue()
+        while not self.empty_queue():
+            startgp, max_cell_counts, cells = self.pop()
+            for cell in self.get_cells_to_try(startgp, max_cell_counts, cells):
+                for gp in startgp.insert_point(cell):
+                    if gp not in self.seen:
+                        self.seen.add(gp)
+                        if (self.satisfies_obstructions(gp) and
```
enadeau: You probably have the answer to this. If a permutation have been yielded then it has been seen. Isn't the yielded check redundant with the seen test?

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -0,0 +1,123 @@
...
+               for max_cell_count in max_cell_counts):
+            return
+        for cell in cells:
+            if any(currcellcounts[cell] < max_cell_count[cell]
+                   for max_cell_count in max_cell_counts):
+                yield cell
+
+    def minimal_gridded_perms(self) -> Iterator[GriddedPerm]:
+        """Yield all minimal gridded perms on the tiling."""
+        if not self.requirements:
```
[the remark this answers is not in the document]
christianbean: Good point

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -0,0 +1,123 @@
...
+        self.obstructions = tiling.obstructions
+        self.requirements = tiling.requirements
+        self.active_cells = union_reduce(union_reduce(gp.pos for gp in req)
+                                         for req in self.requirements)
+        self.seen = set()
+        self.yielded = set()
+        self.queue = []
+        heapify(self.queue)
+
+    def pop(self) -> Info:
```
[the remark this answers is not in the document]
christianbean: This is only because I changed the queue system. I agree I can simplify this.

tilings/algorithms/minimal_gridded_perms.py
christianbean: Much nicer, I'm sold.

tilings/algorithms/minimal_gridded_perms.py
christianbean: A max_cell_count in max_cell_count keeps track of the theoretical bound on the number of points required in each cell in order to create a gridded perm from each requirement list (e.g. if we have R = [[a,b], [c,d]], then there will be a max count for the choices (a, b), (a, c), (b, c), and (b, d).). Therefore this checks that if I am to add another cell, I will definitely be going over the theoretical bound.

tilings/algorithms/minimal_gridded_perms.py
christianbean: Potentially a yielded gridded permutation could be contained in the new gridded perm. As we are not building left to right, but jumping around cells, I can't guarantee there is a unique path to each gridded permutation we come to.

tilings/algorithms/minimal_gridded_perms.py
enadeau: I get it

tilings/algorithms/minimal_gridded_perms.py
enadeau: Ah I did not realize the **sub** part

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -0,0 +1,180 @@
...
+                temp.sort()
+                # update the res
+                new_pos = [(idx[0], val[0]) for idx, val in temp]
+                new_patt = Perm.to_standard(val for _, val in temp)
+                res = GriddedPerm(new_patt, new_pos)
+        return res
+
+    @staticmethod
+    def cell_counter(*gps: GriddedPerm) -> Counter:
+        """Returns a counter of the sum of number of cells used by the gridded
```
enadeau: The docstring is outdated

tilings/tiling.py
```diff
@@ -902,7 +904,8 @@ def is_empty(self) -> bool:
             return True
```
enadeau, line 904: I think those two edge cases are now covered by MinimalGriddedPerms

tilings/tiling.py
```diff
@@ -933,28 +936,28 @@ def gridded_perms(self,
...
-                    maxlen=len(gp1) + len(gp2)))
-        merged_tiling = Tiling(self.obstructions, reqs + [new_req],
-                               remove_empty=remove_empty)
-        return merged_tiling
+        mgps = MinimalGriddedPerms(self)
+        requirements = tuple(Requirement(gp.patt, gp.pos) for gp in
+                             mgps.minimal_gridded_perms())
+        return self.__class__(self.obstructions, (requirements,))
+
+    def minimal_gridded_perms(self, yield_non_minimal=False):
```
enadeau, line 950: Having this flag is kind of weird. I would remove it and change `is_empty` to
```python
MGP = MinimalGriddedPerms(self)
return any(True for _ in MGP.minimal_gridded_perms(yield_non_minimal))
```

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -0,0 +1,447 @@
...
+                mindex, maxdex, minval, maxval = gp.get_bounding_box(cell)
+                # We make sure we only insert to the right of where the last
+                # point was inserted into the cell.
+                mindex = max(mindex, last_mindices[cell])
+                for idx, val in product(range(maxdex, mindex - 1, -1),
+                                        range(minval, maxval + 1)):
+                    nextgp = gp.insert_specific_point(cell, idx, val)
+                    # The work_packets_done is used to ensure the same
+                    # task is not processed twice.
+                    key = (nextgp, gps, next_cell)
```
enadeau, line 417: I would change the line below to. It would be a bit less nested.
```python 
                    if key in self.work_packets_done:
                        continue
                    self.work_packets_done.add(key)
```

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -0,0 +1,447 @@
...
+                    if key not in self.work_packets_done:
+                        self.work_packets_done.add(key)
+                        # Only work with gridded permutations avoiding the
+                        # obstructions, and those which we haven't yielded a
+                        # subgridded permutation.
```
enadeau, line 422: Also here I would try to use an early `continue`

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -0,0 +1,447 @@
...
+        res = self.requirements_up_to_cell.get((cell, gps))
+        if res is None:
+            res = tuple(req.get_gridded_perm_in_cells(
+                                    frozenset(c for c in req.pos if c < cell))
+                        for req in gps)
+            self.requirements_up_to_cell[(cell, gps)] = res
+        return res
+
+    def get_relevant_obstructions(self, gp: GriddedPerm) -> GPTuple:
+        """Get the requirements that involve only cells in the gp."""
```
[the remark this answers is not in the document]
christianbean: Fixed

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -0,0 +1,447 @@
...
+        res = self.relevant_requirements.get(cells)
+        if res is None:
+            res = tuple(tuple(req for req in reqlist
+                              if all(c in cells for c in req.pos))
+                        for reqlist in self.requirements)
+            self.relevant_requirements[cells] = res
+        return res
+
+    def satisfies_obstructions(self, gp: GriddedPerm,
+                               must_contain: Cell = None) -> bool:
```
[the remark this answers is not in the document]
christianbean: Fixed

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -0,0 +1,447 @@
...
+                # update the res
+                new_pos = [(idx[0], val[0]) for idx, val in temp]
+                new_patt = Perm.to_standard(val for _, val in temp)
+                res = GriddedPerm(new_patt, new_pos)
+        return res
+
+    @staticmethod
+    def cell_counter(*gps: GriddedPerm) -> Dict[Cell, int]:
+        """Returns a counter of the sum of number of cells used by the gridded
+        permutations. If 'counter' is given, will update this Counter."""
```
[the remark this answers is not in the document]
christianbean: Fixed

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -0,0 +1,447 @@
...
+        res = self.max_cell_counts.get(gps)
+        if res is None:
+            # we work with the upward closure.
+            gps = self.get_upward_closure(gps)
+            res = MinimalGriddedPerms.cell_counter(*gps)
+            # now we look for any cells containing more than one full req and
+            # are not involved with any other requirements, because we may be
+            # able to apply a better upper bound to the number of points
+            # required in the cell
+            for cell in set(chain.from_iterable(gp.pos for gp in gps)):
```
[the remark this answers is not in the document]
christianbean: Good spot! Fixed.

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -0,0 +1,447 @@
...
+                    # patterns
+                    if len(in_this_cell) >= 2:
+                        in_this_cell = frozenset(in_this_cell)
+                        # the dictionary is patt: v, where v is the number we
+                        # can subtract from the naive bound
+                        res[cell] -= better_bounds.get(in_this_cell, 0)
+            self.max_cell_counts[gps] = res
+        return res
+
+    _better_bounds = None
```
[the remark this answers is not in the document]
christianbean: The logic behind this is that this approach can prevent a long load up time at the start of each run, and it will then only be loaded if necessary.

tilings/tiling.py
christianbean: I agree, but it prevents creating an MGP instance.

tilings/tiling.py
christianbean: Agree, fixed.

tilings/algorithms/minimal_gridded_perms.py
christianbean: Agree, fixed. Also, observed that the attribute was local to this function, so removed it from the class.

tilings/algorithms/minimal_gridded_perms.py
christianbean: Fixed.

christianbean:
> Wow, I'm very impressed by the amount of caching going on in this. So many dictionaries.
In fact the keys are almost always the same ( 5 of the 9 are FrozenSet[Cell], and 2 more are Tuple[Cell, FrozenSet[cell]] ). 
In fact, the following
```python
        self.localised_patts = dict()  # Dict[GPTuple, GPTuple]
        self.max_cell_counts = dict()  # Dict[GPTuple, Dict[Cell, int]]
        self.upward_closures = dict()  # Dict[GPTuple, GPTuple]
        self.known_patts = defaultdict(set)  # DefaultDict[GriddedPerm, Set[GriddedPerm]]
```
are not tiling specific, so could be stored on the class altogether, allowing there corresponding methods to be staticmethods (and in turn also the methods `initial_gp` which relies on `get_upward_closure`).

> Two general comments:
> 
> * I think we should hide the method that are no intended to be used outside the class by prefixing them with a `_`

Agree. I applied this to the methods `get_cells_to_try`, `prepare_queue`. I also moved the function `yielded_subgridded_perm` to be a sub function of `minimal_gridded_perms` as in fact, then the attributes `yielded` and `initial_gps_to_auto_yield` become local to the function, and also this ensures that `minimal_gridded_perms` can be called twice without odd behaviour. 

> * I'm a bit afraid of the `minimal_gridded_perms`. It's very big and deeply nested as pointed out by the linter. I must say that I don't really understand it. I've tried to pin point some easy fix to make it less complex to read but I think it would be very good if we were able to break it into smaller functions.

I removed one layer of complexity by moving the point insertion to a separate function, and use more continues as the specific comments suggested. Do you think it is still a bit monstrous? I think the amount of comments make it look like more.

tilings/algorithms/minimal_gridded_perms.py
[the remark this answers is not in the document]
enadeau: Right, I underestimated the loading time. I thought it would be in nearly instantaneous to import but it actually takes ~1.5s with pypy

tilings/tiling.py
enadeau: Good, not very important anyway

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -301,13 +285,13 @@ def get_upward_closure(self, gps: GPTuple) -> GPTuple:
...
-                         last_cell: Cell,
-                         try_localised: bool
-                        ) -> Iterator[Cell]:
+    def _get_cells_to_try(
+                          self,
+                          gp: GriddedPerm,
+                          gps: GPTuple,
+                          last_cell: Cell,
+                          try_localised: bool
+                         ) -> Iterator[Cell]:
```
enadeau, line 294: The output type is incorrect as it returns a `Tuple[Cell, bool]`

christianbean: Fixed.

tilings/algorithms/minimal_gridded_perms.py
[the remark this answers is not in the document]
jaypantone: If running a whole universe, will it only need to load the first time, or every time?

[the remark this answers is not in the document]
enadeau: Only the first time. It's cached in the class attribute `MinimalGriddedPerms._better_bounds`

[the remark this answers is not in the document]
jaypantone: Okay thanks, so different instances of MinimalGriddedPerms will all be accessing the same copy of the class attribute. Makes sense!

PR #113, opened 2020-04-13
compress using zlib

christianbean:
It saves some space for each stored tiling.


ulfarsson:
This gets lots of failures of Travis, which makes sense because it doesn't know how to read from the database. I think that is the issue... Assuming we want to switch to zlib, what is the best way to move forward? Perhaps do a single pass over the database and change the compression stored for each tiling, by first decompressing with the old version, and then compressing with the new one?

PR #114, opened 2020-04-27
Add fake verification strategy

enadeau:
It's not the first time we have to use this type of strategy. I think it would be good to add it.


enadeau:
We have implement that at the css level

PR #118, opened 2020-05-11
Strategy and class compression

jaypantone:
Adds functions for strategy and class compression.


christianbean:
Moving PR to develop;

PR #120, opened 2020-05-13
Testing tilescope

enadeau:


enadeau:
Before merging this one we need to fix the recursion depth problem as prevents the test from passing on CPython 3.6

PR #122, opened 2020-05-13
image of bijection is Tuple[Optional[GriddedPerm], ...]

christianbean:
This also fixes the fact direction wasn't passed in reg ins encoding strategy

Includes a test for a regular insertion encoding class, sampling, and counting objects


christianbean:
This PR should be addressed after the pull request Fix generation bug #87 is accepted on comb_spec_searcher

PR #124, opened 2020-05-18
init algorithm

christianbean:
This moves some of the code in the initialiser from Tiling to an algorithm class

### DIFF
--- a/tilings/algorithms/gridded_perm_reduction.py
+++ b/tilings/algorithms/gridded_perm_reduction.py
@@ -1,115 +1,38 @@
 from collections import defaultdict
 from functools import partial
-from typing import Iterable, Iterator, List, Optional, Set, Tuple
+from typing import Callable, Dict, Iterable, List, Optional, Set, Tuple
 
 from comb_spec_searcher.utils import cssmethodtimer
+
 from ..griddedperm import GriddedPerm
 
 Cell = Tuple[int, int]
 Requirement = Tuple[GriddedPerm, ...]
 
-func_calls = defaultdict(int)
-func_times = defaultdict(float)
-
-
-class MinimalGriddedPermSet:
-
-    func_calls = func_calls
-    func_times = func_times
-
-    def __init__(
-        self,
-        griddedperms: Iterable[GriddedPerm],
-        minimize: bool = True,
-        sort: bool = True,
-    ):
-        if sort:
-            self.griddedperms: Tuple[GriddedPerm, ...] = tuple(sorted(griddedperms))
-        else:
-            self.griddedperms = tuple(griddedperms)
-        if minimize:
-            self._minimize()
-
-    @cssmethodtimer("MinimalGriddedPermSet._minimize")
-    def _minimize(self) -> None:
-        """
-        Removes non minimal gridded permutations from the set.
-        """
-        res: List[GriddedPerm] = []
-        for gp_to_add in self.griddedperms:
-            if all(gp_already_added not in gp_to_add for gp_already_added in res):
-                res.append(gp_to_add)
-        if len(res) != len(self.griddedperms):
-            # self.griddedperms = tuple(sorted(res))
-            self.casting(res)
-
-    @cssmethodtimer("MinimalGriddedPermSet._recasting")
-    def casting(self, res):
-        self.griddedperms = tuple(sorted(res))
-
-    @cssmethodtimer("MinimalGriddedPermSet.factors")
-    def factors(self) -> Set[GriddedPerm]:
-        if self.griddedperms:
-            res: Set[GriddedPerm] = set(self.griddedperms[0].factors())
-            for gp in self.griddedperms[1:]:
-                if not res:
-                    break
-                res = res.intersection(gp.factors())
-        return res
-
-    @cssmethodtimer("MinimalGriddedPermSet.union_subgps")
-    def union_subgps(self, other: Iterable[GriddedPerm]) -> "MinimalGriddedPermSet":
-        if not isinstance(other, MinimalGriddedPermSet):
-            other = tuple(MinimalGriddedPermSet(other))
-        else:
-            other = tuple(MinimalGriddedPermSet(other))
-        avoiding = tuple(gp for gp in self if gp.avoids(*other))
-        return MinimalGriddedPermSet(avoiding + other, minimize=False)
-
-    def __bool__(self):
-        return bool(self.griddedperms)
-
-    def __eq__(self, other):
-        return self.griddedperms == other.griddedperms
-
-    def __len__(self) -> int:
-        return len(self.griddedperms)
-
-    @cssmethodtimer("MinimalGriddedPermSet.__lt__")
-    def __lt__(self, other):
-        if isinstance(other, MinimalGriddedPermSet):
-            return self.griddedperms < other.griddedperms
-        raise NotImplementedError("Can only compare two MinimalGriddedPermSet")
-
-    def __iter__(self) -> Iterator[GriddedPerm]:
-        return iter(self.griddedperms)
-
-    def __getitem__(self, idx: int):
-        return self.griddedperms[idx]
-
-    def __repr__(self):
-        return self.__class__.__name__ + "({})".format(
-            ", ".join(repr(gp) for gp in self.griddedperms)
-        )
-
-    def __str__(self):
-        return "{{{}}}".format(", ".join(str(gp) for gp in self.griddedperms))
+func_calls: Dict[str, int] = defaultdict(int)
+func_times: Dict[str, float] = defaultdict(float)
 
 
 class GriddedPermReduction:
     func_calls = func_calls
     func_times = func_times
+    passes: Dict[int, int] = defaultdict(int)
 
     @cssmethodtimer("GriddedPermReduction.__init__")
     def __init__(
         self,
-        obstructions: Iterable[GriddedPerm],
-        requirements: Iterable[Iterable[GriddedPerm]],
+        obstructions: Tuple[GriddedPerm, ...],
+        requirements: Tuple[Tuple[GriddedPerm, ...], ...],
+        sorted_input: bool = False,
     ):
-        self._requirements = tuple(
-            sorted(MinimalGriddedPermSet(r) for r in requirements)
-        )
-        self._obstructions = MinimalGriddedPermSet(obstructions)
+        # Only using MGP for typing purposes.
+        if sorted_input:
+            self._obstructions = obstructions
+            self._requirements = requirements
+        else:
+            self._obstructions = tuple(sorted(obstructions))
+            self._requirements = tuple(sorted(tuple(sorted(r)) for r in requirements))
+
         self._minimize_griddedperms()
 
     @property
@@ -119,15 +42,6 @@ class GriddedPermReduction:
     @property
     def requirements(self):
         return tuple(tuple(requirement) for requirement in self._requirements)
-
-    @cssmethodtimer("GriddedPermReduction._clean_isolated")
-    def _clean_isolated(self, obstruction: GriddedPerm) -> GriddedPerm:
-        """Remove the isolated factors that are implied by requirements
-        from all obstructions."""
-        for factor in obstruction.factors():
-            if self._griddedperm_implied_by_some_requirement(factor):
-                obstruction = obstruction.remove_cells(factor.pos)
-        return obstruction
 
     @cssmethodtimer("GriddedPermReduction._minimize_griddedperms")
     def _minimize_griddedperms(self) -> None:
@@ -139,10 +53,14 @@ class GriddedPermReduction:
         """
 
         def set_empty():
-            self._obstructions = MinimalGriddedPermSet([GriddedPerm.empty_perm()])
+            self._obstructions = (GriddedPerm.empty_perm(),)
             self._requirements = tuple()
 
+        self._obstructions = self._minimize(self._obstructions)
+
+        i = 0
         while True:
+            i += 1
             # Minimize the set of obstructions
             minimized_obs = self.minimal_obs()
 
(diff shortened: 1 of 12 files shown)


tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -0,0 +1,264 @@
...
+    def _minimize(self) -> None:
+        """
+        Removes non minimal gridded permutations from the set.
+        """
+        res: List[GriddedPerm] = []
+        for gp_to_add in self.griddedperms:
+            if all(gp_already_added not in gp_to_add for gp_already_added in res):
+                res.append(gp_to_add)
+        self.griddedperms = tuple(sorted(res))
+
```
jaypantone, line 23: We can think about how to optimize this.

tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -0,0 +1,264 @@
...
+        The set of obstructions are first reduced to a minimal set. The
+        requirements that contain any obstructions are removed from their
+        respective lists. If any requirement list is empty, then the tiling is
+        empty.
+        """
+        while True:
+            # Minimize the set of obstructions
+            minimized_obs = MinimalGriddedPermSet(
+                [self._clean_isolated(ob) for ob in self._obstructions]
+            )
```
jaypantone, line 102: bad

tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -0,0 +1,264 @@
...
+                    for f in gp.factors():
+                        # if factor implied by some requirement list then we
+                        # remove it from the gridded perm
+                        if not any(
+                            all(f in g for g in other_requirement)
+                            for other_requirement in requirements
+                            if not all(
+                                any(g2 in g1 for g2 in requirement)
+                                for g1 in other_requirement
+                            )
```
jaypantone, line 185: Hmmmm

tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -0,0 +1,332 @@
...
+            if (
+                self._obstructions == minimized_obs
+                and self._requirements == minimized_requirements
+            ):
+                break
+
+            self._obstructions = minimized_obs
+            self._requirements = minimized_requirements
+        GriddedPermReduction.passes[i] += 1
+
```
jaypantone, line 85: # if there is a requirement for which every component contains the same factor of
    # obstruction, then that factor can be removed from obstruction

tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -0,0 +1,332 @@
...
+                    for factor in factors
+                ):
+                    idx_to_remove.add(i)
+                    break
+        return sorted(
+            requirement
+            for idx, requirement in enumerate(requirements)
+            if idx not in idx_to_remove
+        )
+
```
jaypantone, line 261: # detect if gridded_perm is contained in every gp of requirement

tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -0,0 +1,332 @@
...
+    @cssmethodtimer("GriddedPermReduction._griddedperm_implied_by_requirement")
+    def _griddedperm_implied_by_requirement(
+        self, griddedperm: GriddedPerm, requirement: Iterable[GriddedPerm]
+    ):
+        """
+        Return True, if the containment of the gridded perm is implied by
+        the containment of the requirement.
+        """
+        return all(griddedperm in gp for gp in requirement)
+
```
jaypantone, line 271: # detect if there exists a requirement such that every component in that requirement
    # contains griddedperm

tilings/algorithms/gridded_perm_reduction.py
jaypantone: # uses requirements to shrink obstructions

tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -0,0 +1,332 @@
...
+            partial(self.remove_avoided, obstructions=obstructions),
+            self.remove_redundant_requirements,
+        )
+        minimized_requirements = self._requirements
+        for algo in algos:
+            minimized_requirements = algo(minimized_requirements)
+            if any(not r for r in minimized_requirements):
+                return (tuple(),)
+        return tuple(minimized_requirements)
+
```
jaypantone, line 115:
```
# If all of the gp's in a requirment list contain the same factor, you can remove
# that factor from all of them and add it as it's own size one requirement list
```

tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -0,0 +1,332 @@
...
+                for factor in factors:
+                    res.append((factor,))
+                res.append(
+                    tuple(
+                        gp.get_gridded_perm_in_cells(remaining_cells)
+                        for gp in requirement
+                    )
+                )
+        return res
+
```
jaypantone, line 154:
```
    # if a gridded perm G in a requirment has a factor that appears in EVERY gridded
    # perm of some other requirement, then that factor can be removed from G
```

tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -0,0 +1,332 @@
...
+                                other_requirement, requirement
+                            )
+                        ),
+                    ):
+                        cells.extend(f.pos)
+                newgp = gp.get_gridded_perm_in_cells(cells)
+                newgps.append(newgp)
+            cleaned_reqs.append(tuple(newgps))
+        return cleaned_reqs
+
```
jaypantone, line 188:
```# We can remove any gridded perm in any requiremenet if the gridded perm contains
# an obstruction
```

tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -0,0 +1,332 @@
...
+            cleanreq = tuple(
+                gp for gp in self._minimize(requirement) if gp.avoids(*obstructions)
+            )
+            # If cleanreq is empty, then can not contain this requirement so
+            # the tiling is empty.
+            if not cleanreq:
+                return [tuple()]
+            res.append(cleanreq)
+        return res
+
```
jaypantone, line 212:
```
    # If a requirement contains a gridded perm, each of whose factors is implied by
    # a (possibly different) requirement, then this gridded perm always exists and so
    # the whole requirement can be deleted.
```

tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -0,0 +1,305 @@
...
+            if not all(requirement):
+                continue
+            newgps: List[GriddedPerm] = []
+            for gp in requirement:
+                cells: List[Cell] = []
+                for f in gp.factors():
+                    if not self._griddedperm_implied_by_some_requirement(
+                        f,
+                        (
+                            other_requirement
```
enadeau, line 204: Can we just replace this big thing by
`requirements[:idx] + requirements[idx+1:]`

enadeau: Or if your concern about creating new list we can also use itertools.islice

tilings/tiling.py
```diff
@@ -64,16 +67,31 @@ class Tiling(CombinatorialClass):
...
+        remove_empty_rows_and_cols: bool = True,
         derive_empty: bool = True,
-        minimize: bool = True,
+        simplify: bool = True,
         sorted_input: bool = False,
+        already_minimized_obs: bool = False,
     ) -> None:
+        """
+        - if not sorted_input, input will be sorted
+        - if minimize, will be minimized
```
enadeau, line 87: This doctrstring does not match the actual arguments of the method

tilings/tiling.py
```diff
@@ -125,37 +145,38 @@ def from_perms(
...
     def _fill_empty(self) -> None:
         """Add size one obstructions to cells that are empty."""
         add = []
         (i, j) = self.dimensions
+        active_cells = self.active_cells
+        empty_cells = self.empty_cells
         for x in range(i):
             for y in range(j):
-                if (x, y) not in self.active_cells and (x, y) not in self.empty_cells:
+                if (x, y) not in active_cells and (x, y) not in empty_cells:
```
enadeau, line 157: I would replace this big nested statement by
```python
all_cells = set(itertools.product(*map(range, dimensions)))
add = {GriddedPerm.single_cell(Perm((0,)), cell) for cell in all_cells - active_cells - empty_cells}
```
but maybe I'm just obsessed about avoiding big nesting
```

File: tilings/tiling.py [hunk 14]
```diff
@@ -928,8 +811,11 @@ def sub_tiling(self, cells: Iterable[Cell], factors: bool = False) -> "Tiling":
             if (factors and ass.gps[0].pos[0] in cells)
             or all(c in cells for c in chain.from_iterable(gp.pos for gp in ass.gps))
         )
-        return self.__class__(obstructions, requirements, assumptions)
+        return self.__class__(
+            obstructions, requirements, assumptions, simplify=False, sorted_input=True
+        )
 
+    @cssmethodtimer("Tiling.find_factors")
```
enadeau, line 818: Why do we time the `find_factors` methods?

tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -0,0 +1,305 @@
...
+            minimized_requirements = self.minimal_reqs(minimized_obs)
+            minimized_requirements = tuple(
+                sorted(tuple(sorted(set(req))) for req in minimized_requirements)
+            )
+
+            if minimized_requirements and not minimized_requirements[0]:
+                set_empty()
+                break
+
+            if self._requirements == minimized_requirements:
```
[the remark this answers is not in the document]
jaypantone: No, this is one of the improvements I made. The `minimal_obs` function uses the requirements to simplify obstructions (and is idempotent). So if, in one pass, the obstructions change but the requirements don't, there is still no need to loop again.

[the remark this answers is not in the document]
jaypantone: @christianbean Do you agree with this logic?

tilings/algorithms/gridded_perm_reduction.py
jaypantone: I'm not sure I understand this comment. This function shrinks GPs in requirements by removing some factors.

tilings/tiling.py
jaypantone: Do you claim your version is faster? I find it definitely less readable, but maybe that's worth it if there's a speed advantage.

tilings/tiling.py
jaypantone: In case it's slow? We can probably remove this.

tilings/algorithms/gridded_perm_reduction.py
enadeau: I meant that the second argument of `_griddedperm_implied_by_some_requirement` could be replaced  by this which is much shorted.

tilings/tiling.py
jaypantone: Fixed.

tilings/tiling.py
enadeau: Not claiming that it's faster only that it is more readable. Probably not worth changing if you don't find it more readable

tilings/algorithms/gridded_perm_reduction.py
jaypantone: Oh, I see. I think itertools.islice would be okay. @christianbean?

tilings/tiling.py
enadeau: My question came from the fact that it's the only method not related to the initializer that we are timing. We might have added the timing by accident.

jaypantone: I think it was used at one point, but maybe not anymore. We should remove it.

tilings/algorithms/gridded_perm_reduction.py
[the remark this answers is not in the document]
enadeau: Took a closer look. I agree with the logic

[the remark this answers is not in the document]
christianbean: I agree too

tilings/algorithms/gridded_perm_reduction.py
christianbean: That is the best option

tilings/tiling.py
jaypantone: Fixed. Added timing to Tiling._simplify_griddedperms.

tilings/algorithms/gridded_perm_reduction.py
jaypantone: Fixed.

tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -0,0 +1,303 @@
...
+            GriddedPermReduction.factored_reqs,
+            self.cleaned_requirements,
+            partial(self.remove_avoided, obstructions=obstructions),
+        )
+        minimized_requirements = self._requirements
+        for algo in algos:
+            minimized_requirements = algo(minimized_requirements)
+            if any(not r for r in minimized_requirements):
+                return (tuple(),)
+        return tuple(sorted(minimized_requirements))
```
christianbean, line 138: I think this sort is redundant

tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -0,0 +1,303 @@
...
+            newgps: List[GriddedPerm] = []
+            for gp in requirement:
+                cells: List[Cell] = []
+                for f in gp.factors():
+                    if not self._griddedperm_implied_by_some_requirement(
+                        f,
+                        itertools.chain(
+                            itertools.islice(requirements, idx),
+                            itertools.islice(requirements, idx + 1, None),
+                        ),
```
christianbean, line 207: we should just import chain and islice

tilings/tiling.py
```diff
@@ -35,13 +36,15 @@
...
     Fusion,
+    GriddedPermReduction,
     GriddedPermsOnTiling,
     MinimalGriddedPerms,
     ObstructionTransitivity,
     RequirementPlacement,
     RowColSeparation,
     SubobstructionInferral,
 )
+from .algorithms.gridded_perm_reduction import func_calls, func_times
```
christianbean, line 47: probably makes more sense for these to be in tiling.py file when fist defined

christianbean: this leads to circular imports so ignore this for now

PR #125, opened 2020-05-18
Assumptions

christianbean:
Added an AbstractAssumtpion class.

In particular a `TrackingAssumption`, which will keep track of a set of gridded permutations. 

- [x] basic methods: (to_bytes, to_jsonable, initialiser, __eq__, etc) methods 
- [x] update functions returning Tiling to pass on assumptions
- [x] update strategies to pass on Assumptions when a strategy is applied in particular, TrackingAssumptions should be stretched where necessary
- [x] think about future proofing is-empty, it should consider assumptions (EDIT: decision was made to not plan for ideas we don't have yet, and are not sure what they will look like)

### DIFF
--- a/tilings/algorithms/factor.py
+++ b/tilings/algorithms/factor.py
@@ -74,9 +74,6 @@ class Factor:
         For each TrackingAssumption unite all the positions of the gridded perms.
         """
         for ass in self._tiling.assumptions:
-            assert isinstance(
-                ass, TrackingAssumption
-            ), "not implemented factor for assumption given"
             ass_cells = chain.from_iterable(gp.pos for gp in ass.gps)
             self._unite_cells(ass_cells)
 
@@ -161,14 +158,10 @@ class Factor:
             requirements = tuple(
                 req for req in self._tiling.requirements if req[0].pos[0] in component
             )
-            assert all(
-                isinstance(ass, TrackingAssumption) for ass in self._tiling.assumptions
-            ), "factors only implemented for TrackingAssumption"
             assumptions = tuple(
                 ass
                 for ass in self._tiling.assumptions
-                if isinstance(ass, TrackingAssumption)
-                and ass.gps[0].pos[0] in component
+                if ass.gps[0].pos[0] in component
             )
             factors.append((obstructions, requirements, assumptions))
         self._factors_obs_and_reqs = factors
--- a/tilings/algorithms/factor.py
+++ b/tilings/algorithms/factor.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING, Dict, Iterable, Iterator, List, Optional, Set,
 
 from permuta.misc import UnionFind
 from tilings import GriddedPerm
+from tilings.assumptions import TrackingAssumption
 from tilings.misc import partitions_iterator
 
 if TYPE_CHECKING:
@@ -30,7 +31,13 @@ class Factor:
         self._cell_unionfind = UnionFind(nrow * ncol)
         self._components: Optional[Tuple[Set[Cell], ...]] = None
         self._factors_obs_and_reqs: Optional[
-            List[Tuple[Tuple[GriddedPerm, ...], Tuple[ReqList, ...]]]
+            List[
+                Tuple[
+                    Tuple[GriddedPerm, ...],
+                    Tuple[ReqList, ...],
+                    Tuple[TrackingAssumption, ...],
+                ]
+            ]
         ] = None
 
     def _cell_to_int(self, cell: Cell) -> int:
@@ -58,6 +65,19 @@ class Factor:
         for c2 in cell_iterator:
             c2_int = self._cell_to_int(c2)
             self._cell_unionfind.unite(c1_int, c2_int)
+
+    def _unite_assumptions(self) -> None:
+        """
+        For each TrackingAssumption unite all the positions of the gridded perms.
+        """
+        for ass in self._tiling.assumptions:
+            if isinstance(ass, TrackingAssumption):
+                for gp in ass.gps:
+                    self._unite_cells(gp.pos)
+            else:
+                raise NotImplementedError(
+                    "Not implemented factors for assumption {}".format(ass)
+                )
 
     def _unite_obstructions(self) -> None:
         """
@@ -120,7 +140,11 @@ class Factor:
 
     def _get_factors_obs_and_reqs(
         self,
-    ) -> List[Tuple[Tuple[GriddedPerm, ...], Tuple[ReqList, ...]]]:
+    ) -> List[
+        Tuple[
+            Tuple[GriddedPerm, ...], Tuple[ReqList, ...], Tuple[TrackingAssumption, ...]
+        ],
+    ]:
         """
         Returns a list of all the irreducible factors of the tiling.
         Each factor is a tuple (obstructions, requirements)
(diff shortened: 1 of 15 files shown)


tilings/tiling.py
```diff
@@ -1185,6 +1276,13 @@ def requirements(self) -> Tuple[Tuple[GriddedPerm, ...], ...]:
     def total_requirements(self) -> int:
         return len(self._requirements)
 
+    @property
+    def assumptions(self) -> Tuple[AbstractAssumption, ...]:
+        return self._assumptions
+
+    def total_assumptions(self) -> int:
```
enadeau, line 1283: Maybe `num_assumptions` would be a better name?

tilings/tiling.py
```diff
@@ -1223,6 +1321,19 @@ def dimensions(self) -> Tuple[int, int]:
                 self._dimensions = (max(rows) + 1, max(cols) + 1)
         return self._dimensions
 
+    def underlying_tiling(self):
```
enadeau, line 1324: Maybe `remove_assumption` would be a better name?

tilings/algorithms/fusion.py
```diff
@@ -113,6 +116,23 @@ def requirements_fuse_counters(self):
         self._requirements_fuse_counters = counters
         return self._requirements_fuse_counters
 
+    @property
+    def assumptions_fuse_counters(self):
+        """
+        List of fuse counters for each of the TrackedAssumptions of the tiling.
+        """
+        if self._assumptions_fuse_counters is not None:
+            return self._requirements_fuse_counters
```
enadeau, line 125: I guess you mean `return  self._assumptions_fuse_counter`

tilings/algorithms/factor.py
```diff
@@ -120,7 +139,11 @@ def get_components(self) -> Tuple[Set[Cell], ...]:
 
     def _get_factors_obs_and_reqs(
         self,
-    ) -> List[Tuple[Tuple[GriddedPerm, ...], Tuple[ReqList, ...]]]:
+    ) -> List[
+        Tuple[
+            Tuple[GriddedPerm, ...], Tuple[ReqList, ...], Tuple[TrackingAssumption, ...]
+        ],
+    ]:
```
enadeau, line 146: The docstring need to be updated to the new output

tilings/algorithms/factor.py
```diff
@@ -254,5 +262,6 @@ def _unite_all(self) -> None:
         """
         Unite all the cells that share an obstruction or a requirement list.
         """
+        self._unite_assumptions()
```
enadeau, line 265: Not sure how the Interleaving and MonotoneInterleaving factor handle assumption. Maybe we should just raise an error if the tiling as assumptions in those cases?

tilings/assumptions.py
```diff
@@ -0,0 +1,89 @@
...
+    @abc.abstractmethod
+    def __hash__(self) -> int:
+        pass
+
+    def __eq__(self, other) -> bool:
+        return isinstance(self, other.__class__) and self.__dict__ == other.__dict__
+
+    @abc.abstractmethod
+    def __lt__(self, other) -> bool:
+        return bool(self.__class__.__name__ < other.__class__.__name__)
```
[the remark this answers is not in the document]
christianbean: yes

tilings/assumptions.py
```diff
@@ -0,0 +1,89 @@
...
+            module, d.pop("assumption_class")
+        )
+        return AssumptionClass.from_dict(d)
+
+    @abc.abstractmethod
+    def __hash__(self) -> int:
+        pass
+
+    def __eq__(self, other) -> bool:
+        return isinstance(self, other.__class__) and self.__dict__ == other.__dict__
```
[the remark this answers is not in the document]
christianbean: agree

tilings/strategies/assumption_splitting.py
```diff
@@ -0,0 +1,136 @@
...
+            split_gps: List[List[GriddedPerm]] = [[] for _ in range(len(components))]
+            for gp in ass.gps:
+                for idx, component in enumerate(components):
+                    if all(cell in component for cell in gp.pos):
+                        split_gps[idx].append(gp)
+                        # only add to one list
+                        break
+                else:
+                    # gridded perms can't be partitioned
+                    new_assumptions.extend([ass])
```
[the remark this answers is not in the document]
christianbean: The purpose of this function is to look at each assumptions, and partition them into "sub"  assumptions, according to the factors: 

i.e. if the assumption is {1: (0, 0), 1: (1, 1)} then this will be turned into {1: (0, 0)} and {1: (1, 1)}. This will allow us to factor, as the original assumption will forbid the separating (0, 0) and (1, 1). 

> should consider the factor of the gp

Not in this function, but this should be a separate strategy, as the way to count it will be different.

tilings/tiling.py
christianbean: I used the word total, as other methods used this, e.g. total_obstructions, total_requirements, etc. I can change it if you like

tilings/algorithms/factor.py
christianbean: They should behave the same, although when using it, it could be beneficial for example, to have the splitting strategy use the relevant factor algorithm.

tilings/tiling.py
christianbean: sure

christianbean:
> At the moment we define an `AbstractAssumption` and `TrackingAssumption`. Everywhere in the code we just assert that our assumption is a `TrackingAssumptions`, we're not using any of the power of having this general framework. As we add new types of assumptions, I'm scare that we just gone blow up the size of our function with special casing for every type of assumption. What do you think?

I tend to agree, but the issue is I don't know what future assumptions will look like. Therefore would rather abstract when that becomes clearer. 
> For example assumption can define a `map` method that map the assumption according to a cell map and return the new assumption. This will raise a non-implemented error for assumption where it's not defined. Then you don't need the check the type of the assumption every time you want to map it.

The map methods needed are always specific to each strategy, so I feel it makes more sense to implement them in the actual strategy file. As we start creating more assumption types it might become clearer what we want to abstract.


> Also would you be against having `Tiling.add_assumptions` method ? I think it would be useful when debugging and working in the interpreter.

Sure, I'll add it.

tilings/tiling.py
enadeau: Ok I didn't knew it was consistent with the other. Probably not worth changing then

christianbean:
The latest commit removes the AbstractAssumption class, this in turn resolves the questions about the bytes methods.

tilings/assumptions.py
```diff
@@ -64,24 +20,33 @@ def avoiding(self, obstructions: Iterable[GriddedPerm]) -> "TrackingAssumption":
...
     def from_dict(cls, d: dict) -> "TrackingAssumption":
         gps = [GriddedPerm.from_dict(gp) for gp in d["gps"]]
         return cls(gps)
 
-    def __lt__(self, other):
-        if isinstance(other, self.__class__):
-            return self.gps < other.gps
-        return super().__lt__(other)
+    def __eq__(self, other) -> bool:
+        try:
```
enadeau, line 32: Thie `__eq__` implementations do not follow the python standard for implementation of binary special method. It is recommended to return the `NotImplemented` constant when the operation is not implemented for a type.

```python
def __eq__(self, other: object) -> bool:
    if not isinstance(other, TrackingAssumption):
        return NotImplemented
    return self.gps == other.gps
```
[https://docs.python.org/3/library/constants.html](https://docs.python.org/3/library/constants.html)

tilings/assumptions.py
```diff
@@ -64,24 +20,33 @@ def avoiding(self, obstructions: Iterable[GriddedPerm]) -> "TrackingAssumption":
...
+            return bool(self.gps == other.gps)
+        except AttributeError:
+            raise NotImplementedError(
+                "Can't compare type {} to {}".format(
+                    self.__class__.__name__, other.__class__.__name__
+                )
+            )
+
+    def __lt__(self, other) -> bool:
```
enadeau, line 41: same as __eq__

tilings/assumptions.py
christianbean: I'll fix this

PR #131, opened 2020-05-26
fix a bug in the ordering of the mgps

enadeau:
Sometime some the Minimal Gridded Perm algorithm was not returning the
gps in increasing length order. This is now fixed

### DIFF
--- a/tilings/algorithms/minimal_gridded_perms.py
+++ b/tilings/algorithms/minimal_gridded_perms.py
@@ -404,18 +404,18 @@ class MinimalGriddedPerms:
         queue: List[QueuePacket] = []
         heapify(queue)
 
-        initial_gps_to_auto_yield: Set[GriddedPerm] = set()
+        initial_gps_to_auto_yield: Dict[int, Set[GriddedPerm]] = defaultdict(set)
         yielded: Set[GriddedPerm] = set()
         for gp in self._prepare_queue(queue):
             if yield_non_minimal:
                 yielded.add(gp)
                 yield gp
             else:
-                initial_gps_to_auto_yield.add(gp)
+                initial_gps_to_auto_yield[len(gp)].add(gp)
 
         def yielded_subgridded_perm(gp: GriddedPerm) -> bool:
             """Return True if a subgridded perm was yielded."""
-            return gp.contains(*yielded, *initial_gps_to_auto_yield)
+            return gp.contains(*yielded, *initial_gps_to_auto_yield[len(gp)])
 
         def _process_work_packet(
             qpacket: QueuePacket, queue: List[QueuePacket],
@@ -477,6 +477,7 @@ class MinimalGriddedPerms:
                         )
 
         work_packets_done: Set[WorkPackets] = set()
+        curr_len = 0
         while queue:
             # take the next gridded permutation of the queue, together with the
             # theoretical counts to create a gridded permutation containing
@@ -485,9 +486,14 @@ class MinimalGriddedPerms:
             # if gp was one of the initial_gps that satisfied obs/reqs, but
             # we weren't sure at the time if it was minimal, then now is the
             # time to check and yield
-            if qpacket.gp in initial_gps_to_auto_yield:
+            if curr_len != len(qpacket.gp):
+                curr_len = len(qpacket.gp)
+                for gp in initial_gps_to_auto_yield[curr_len]:
+                    yielded.add(gp)
+                    yield gp
+            if qpacket.gp in initial_gps_to_auto_yield[curr_len]:
                 # removing this speed up a later check by a tiny bit
-                initial_gps_to_auto_yield.remove(qpacket.gp)
+                initial_gps_to_auto_yield[curr_len].remove(qpacket.gp)
                 if not yielded_subgridded_perm(qpacket.gp):
                     yielded.add(qpacket.gp)
                     yield qpacket.gp
--- a/tilings/algorithms/minimal_gridded_perms.py
+++ b/tilings/algorithms/minimal_gridded_perms.py
@@ -414,7 +414,7 @@ class MinimalGriddedPerms:
 
         def yielded_subgridded_perm(gp: GriddedPerm) -> bool:
             """Return True if a subgridded perm was yielded."""
-            return gp.contains(*yielded, *initial_gps_to_auto_yield[len(gp)])
+            return gp.contains(*yielded)
 
         def _process_work_packet(
             qpacket: QueuePacket, queue: List[QueuePacket],
--- a/tilings/algorithms/minimal_gridded_perms.py
+++ b/tilings/algorithms/minimal_gridded_perms.py
@@ -489,8 +489,9 @@ class MinimalGriddedPerms:
             if curr_len != len(qpacket.gp):
                 curr_len = len(qpacket.gp)
                 for gp in initial_gps_to_auto_yield[curr_len]:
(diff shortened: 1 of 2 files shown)


tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -485,9 +486,14 @@ def _process_work_packet(
             # if gp was one of the initial_gps that satisfied obs/reqs, but
             # we weren't sure at the time if it was minimal, then now is the
             # time to check and yield
-            if qpacket.gp in initial_gps_to_auto_yield:
+            if curr_len != len(qpacket.gp):
+                curr_len = len(qpacket.gp)
+                for gp in initial_gps_to_auto_yield[curr_len]:
+                    yielded.add(gp)
+                    yield gp
```
christianbean, line 493: you need to double check gp is minimal here
```python
if not yielded_subgridded_perm(gp):
    yielded.add(gp)
    yield gp
```

enadeau: True just fixed it

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -485,9 +485,15 @@ def _process_work_packet(
             # if gp was one of the initial_gps that satisfied obs/reqs, but
             # we weren't sure at the time if it was minimal, then now is the
             # time to check and yield
-            if qpacket.gp in initial_gps_to_auto_yield:
+            if curr_len != len(qpacket.gp):
+                curr_len = len(qpacket.gp)
+                for gp in initial_gps_to_auto_yield[curr_len]:
+                    if not yielded_subgridded_perm(qpacket.gp):
```
enadeau, line 491: This is wrong I should actually check `gp` not `qpacket.gp`

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -395,27 +395,26 @@ def minimal_gridded_perms(
...
                 yielded.add(gp)
                 yield gp
             else:
-                initial_gps_to_auto_yield.add(gp)
+                initial_gps_to_auto_yield[len(gp)].add(gp)
 
         def yielded_subgridded_perm(gp: GriddedPerm) -> bool:
             """Return True if a subgridded perm was yielded."""
-            return gp.contains(*yielded, *initial_gps_to_auto_yield)
+            return gp.contains(*yielded, *initial_gps_to_auto_yield[len(gp)])
```
christianbean, line 417: you need to check for all sizes in range(curr_len + 1, len(gp) + 1)

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -485,9 +485,15 @@ def _process_work_packet(
             # if gp was one of the initial_gps that satisfied obs/reqs, but
             # we weren't sure at the time if it was minimal, then now is the
             # time to check and yield
-            if qpacket.gp in initial_gps_to_auto_yield:
+            if curr_len != len(qpacket.gp):
+                curr_len = len(qpacket.gp)
+                for gp in initial_gps_to_auto_yield[curr_len]:
```
christianbean, line 490: you should yield from range(prev_len + 1, curr_len + 1)

tilings/algorithms/minimal_gridded_perms.py
```diff
@@ -485,9 +485,15 @@ def _process_work_packet(
...
+                    if not yielded_subgridded_perm(qpacket.gp):
+                        yielded.add(gp)
+                        yield gp
+            if qpacket.gp in initial_gps_to_auto_yield[curr_len]:
                 # removing this speed up a later check by a tiny bit
-                initial_gps_to_auto_yield.remove(qpacket.gp)
+                initial_gps_to_auto_yield[curr_len].remove(qpacket.gp)
```
christianbean, line 496: this feels unnecessary - you could mark each length as empty after yielding from it

Note, qpacket.gp should never be an initial_gps if you put that continue in the prepare queue method

tilings/algorithms/minimal_gridded_perms.py
enadeau: You're right. I changed it.

tilings/algorithms/minimal_gridded_perms.py
enadeau: Actually I think I only need to check for containment in the yielded gridded perms. It's what I did and it seems to work

enadeau: I think you don't need check anything in the initial in fact. If they've already been yielded then they are in the yielded set and the function already returns true. At least I did it that way and it works.

tilings/algorithms/minimal_gridded_perms.py
enadeau: I added the `continue` we discussed and removed the if statement completely.

enadeau:
I made most of the requested change. The only point I did not understand was about the `yielded_subgridded_perm` function. I left more details on the specific discussion.

tilings/algorithms/minimal_gridded_perms.py
christianbean: This is a good observation - they are all added to the yielded set the moment the first one of that length is found. Hence for gp, yielded contains all initial_gps of length <= len(gp)

enadeau: I don't see it as a problem it will just yield the one point extension and then when it changes length it won't yield the one from `initial_gp_to_auto_yield` since it is already in yielded.

Also, I'm not sure I understand everything about this 1 point extension but I'm not sure I understand how it's not breaking the fact that the permutation are yielded in increasing length?

christianbean: You are right, the order is preserved as the point extension would be yielded first, good point!

I am going to try abstract a bit to make the second question super clear:
- the prepare queue finds some gps, call it S that we know satisfy the obs and reqs
- we pop a gridded perm gp from the queue, 
- we know at this point that all from S with length <= len(gp) are yielded. 
- we then try yield the one point extensions of gp, each of length len(gp) + 1. 
- and repeat

If a point extension isn't a minimal gridded perm, it is added to the priority queue (sorted by length), this is why we ensure they are yielded in increasing length.

enadeau: Ok I think I understand. Any gp on the queue does not satisfy the requirements. Hence, when `len(qpacket.gp) == n`, we are actually yielding permutation of length n+1.

Am I right to think that we should therefore yield the initial_to_auto_yield first by changing line 493-496 from
```python
             initial_to_yield = chain.from_iterable(
                if not yielded_subgridded_perm(qpacket.gp):
                    initial_gps_to_auto_yield.pop(i, tuple())
                    yielded.add(qpacket.gp)
                    for i in range(curr_len + 1, new_len + 1)
                    yield qpacket.gp
                )
```
to
```python
             initial_to_yield = chain.from_iterable(
                if not yielded_subgridded_perm(qpacket.gp):
                    initial_gps_to_auto_yield.pop(i, tuple())
                    yielded.add(qpacket.gp)
                    for i in range(curr_len + 1, new_len + 2)
                    yield qpacket.gp
                )
```

enadeau: I tested it locally and it passes all the tests

christianbean: I agree with this change

PR #133, opened 2020-05-28
No factor in expansions

enadeau:
I've added the no factors option a to the expansion pack and set up a test to check that. However the test is failing because with cPython it does not expand for some reason.

### DIFF
--- a/tests/test_tilescope.py
+++ b/tests/test_tilescope.py
@@ -2,6 +2,7 @@ import pytest
 import sympy
 
 from comb_spec_searcher import CombinatorialSpecification
+from comb_spec_searcher.strategies.rule import VerificationRule
 from comb_spec_searcher.utils import taylor_expand
 from permuta import Perm
 from tilings import GriddedPerm, Tiling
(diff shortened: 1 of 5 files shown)


tests/test_tilescope.py
```diff
@@ -250,3 +251,22 @@ def test_from_tiling():
...
+    pack = TileScopePack.only_root_placements(3, 1)
+    css = TileScope("132", pack)
+    spec = css.auto_search(smallest=True)
+    for comb_class, rule in spec.rules_dict.items():
+        if isinstance(rule, VerificationRule):
+            print(rule)
+            assert comb_class.dimensions == (1, 1)
+            assert comb_class.obstructions != (
+                GriddedPerm(Perm((0, 2, 1)), ((0, 0),) * 3),
+            )
```
christianbean, line 272: I disagree with this as a test. 

The specification itself when finding the sub spec for the verified factorable thing, doesn't need to worry about the root. I'm confident that the equations we write will contain a minus (so no good for sampling) but will give equations that are solvable.  

If we are this scared of doing that then we should forbid verifying any object which contains the basis in any cell.

enadeau: I also think it's safe, my only concern is that when we have the 1x1 database setup we might end up just getting term from the database but we can consider that when it happens. I'll remove the second assert

christianbean: Ok, cool. If you look at line 205 of specification.py, you see funcs is passed the verification rule, which passes the to the verification strategy get genf method. Funcs  contains the functions of the lhs in the specification. 
Then in OneByOneVerificationStrategy, the get genf method passes the funcs to LocalEnumeration get genf, then line 89 in enumeration.py, will then return the relevant generating function (note its a recursive function). Hopefully that makes you a little less worried. 

[Note: perhaps, line 89 should come after line 77, but I decided it would always try remove reqs first]

PR #135, opened 2020-05-29
update only_root_placements

enadeau:


enadeau:
This is the change we discussed in meeting so merging

PR #138, opened 2020-05-31
Computing and Caching Tiling properties

jaypantone:
Making this PR so we can discuss this in our next meeting. Overall it seems slower, which is surprising to me. But I think parts of it are still useful.

### DIFF
--- a/tilings/tiling.py
+++ b/tilings/tiling.py
@@ -3,6 +3,7 @@
 # pylint: disable=too-many-return-statements,too-many-statements
 # pylint: disable=import-outside-toplevel
 import json
+import traceback
 from array import array
 from collections import Counter, defaultdict
 from functools import partial
@@ -19,7 +20,6 @@ from typing import (
     Set,
     Tuple,
 )
-import traceback
 
 import sympy
 
--- a/tilings/tiling.py
+++ b/tilings/tiling.py
@@ -142,6 +142,7 @@ class Tiling(CombinatorialClass):
             # obstructions
             if derive_empty:
                 self._fill_empty()
+                del self._empty_cells
 
     @classmethod
     def from_perms(
--- a/tilings/tiling.py
+++ b/tilings/tiling.py
@@ -128,6 +128,11 @@ class Tiling(CombinatorialClass):
 
         if not any(ob.is_empty() for ob in self.obstructions):
 
+            # Set self._active_cells, self._empty_cells, and self._dimensions
+            # Should eliminate the need for _fill_empty
+            # The function _remove_empty_rows_and_cols is responsible for adjusting
+            self._prepare_properties()
+
             # Remove gridded perms that avoid obstructions from assumptions
             if simplify:
                 self._clean_assumptions()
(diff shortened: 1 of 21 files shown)


tilings/tiling.py
```diff
@@ -56,6 +59,27 @@
 Cell = Tuple[int, int]
 ReqList = Tuple[GriddedPerm, ...]
 
+CellBasis = Dict[Cell, Tuple[List[Perm], List[Perm]]]
+Map = Dict[Cell, Cell]
```
enadeau, line 63: Should we call that `CellMap` to be more explicit?

tilings/tiling.py
```diff
@@ -145,18 +169,53 @@ def from_perms(
...
+                    break
+                index += 1
+            # Now the last point obstruction is at index [index-1]
+            # point_obstructions = self._obstructions[:index]
+            non_point_obstructions = self._obstructions[index:]
+
+            new_point_obstructions = tuple(
+                GriddedPerm(Perm((0,)), (cell,)) for cell in empty_cells
+            )
+            self._obstructions = new_point_obstructions + non_point_obstructions
```
enadeau, line 214: Don't we need to ensure proper sorting of the point obstruction here?

tilings/tiling.py
jaypantone: Agree.

tilings/tiling.py
jaypantone: Credit to @christianbean for this trick. `empty_cells` is already sorted lexicographically (because it's a sub-tuple of itertools.product, which is already sorted lexicographically).

tilings/tiling.py
```diff
@@ -145,18 +169,53 @@ def from_perms(
...
+        # If the first obstruction is the empty perm, we shouldn't do any of this.
+        if len(self._obstructions) > 0 and len(self._obstructions[0]) > 0:
+            # We can assume that self._obstructions is sorted at this point, so to
+            #   extract the point obstructions, we just pass though them until we've
+            #   found the last one, then we slice the list there.
+            index = 0
+            for ob in self._obstructions:
+                if len(ob) > 1:
+                    break
+                index += 1
```
christianbean: We should use the next function for this:
```python
index = next(i for i, ob in enumerate(self._obstructions) if len(ob) > 1)
```
I'll make this change

christianbean: Is it fair to assume that there is at least one of size > 1 ?

tilings/tiling.py
```diff
@@ -145,22 +169,56 @@ def from_perms(
...
+
+        # If the first obstruction is the empty perm, we shouldn't do any of this.
+        if len(self._obstructions) > 0 and len(self._obstructions[0]) > 0:
+            # We can assume that self._obstructions is sorted at this point, so to
+            #   extract the point obstructions, we just pass though them until we've
+            #   found the last one, then we slice the list there.
+            index = next(
+                (i for i, ob in enumerate(self._obstructions) if len(ob) > 1),
+                len(self._obstructions),  # default value
+            )
```
christianbean, line 205: I changed this to use next, rather than the for loop.

tilings/tiling.py
christianbean: the answer was no, needed to add default value being the length of the obstructions

jaypantone: I was worried a solution with enumerate would be slower, and this is one of those things we do millions of times. And yes, you need a default for the case where all obstructions are point obstructions (e.g., the empty tiling).

PR #139, opened 2020-06-01
Enumerating fusion

christianbean:
The PR relies on branch extra-variables of CSS. 

It includes code to enumerate any tree found with tracked fusion.


tilings/strategies/verification.py
```diff
@@ -32,10 +38,33 @@
 ]
 
 
-BasicVerificationStrategy = AtomStrategy
 TileScopeVerificationStrategy = VerificationStrategy[Tiling, GriddedPerm]
 
 
+class BasicVerificationStrategy(AtomStrategy):
```
enadeau, line 44: It feels like something that belongs to css instead of tilings.

tilings/tiling.py
```diff
@@ -181,7 +181,9 @@ def _remove_empty_rows_and_cols(self) -> None:
         """Remove empty rows and columns."""
         # Produce the mapping between the two tilings
         if not self.active_cells:
-            self._forward_map: Dict[Cell, Cell] = {}
+            self._forward_map: Dict[Cell, Cell] = {
```
enadeau, line 184: Why do we need to map the empty cell?

tilings/strategies/verification.py
```diff
@@ -32,10 +38,33 @@
...
+        """
+        Verification strategies must contain a method to count the objects.
+        """
+        if not isinstance(comb_class, Tiling):
+            raise NotImplementedError
+        cast(Tiling, comb_class)
+        gp = next(comb_class.minimal_gridded_perms())
+        expected = {"n": len(gp)}
+        for assumption in comb_class.assumptions:
+            count = len(
```
enadeau, line 58: I think this is a method we want on TrackingAssumption. It feels like we'll need to do that often

tilings/strategies/assumption_splitting.py
```diff
@@ -30,7 +35,72 @@ def reliance_profile(self, n: int, **parameters: int) -> RelianceProfile:
...
+
+        # TODO: this should take into consideration the reliance profile.
+        """
+        rec = subrecs[0]
+        res = 0
+        for sub_params in self._valid_compositions(**parameters):
+            res += rec(n, **sub_params)
+        return res
+
+    def _valid_compositions(self, **parameters: int) -> Iterator[Dict[str, int]]:
```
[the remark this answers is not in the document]
christianbean: This function is the one that need most updating but thought I'd optimise later. It should be more like the valid_compositions method in CartesianProduct on css. 

I'll add a docstring for its current functionality.

tilings/strategies/assumption_splitting.py
```diff
@@ -30,7 +35,72 @@ def reliance_profile(self, n: int, **parameters: int) -> RelianceProfile:
...
+        """
+        TODO: this should consider reliance profiles, and when variables are
+        sub variable of other variables.
+        """
+
+        def compositions(value: int, parameters: Tuple[str, ...]):
+            def _compositions(value: int, length: int) -> Iterator[Tuple[int, ...]]:
```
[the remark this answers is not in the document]
christianbean: I agree, I should do that.

tilings/strategies/assumption_splitting.py
```diff
@@ -60,37 +130,52 @@ class SplittingStrategy(Strategy[Tiling, GriddedPerm]):
...
-                    break
-            else:
-                new_assumptions.extend(
-                    [TrackingAssumption(gps) for gps in split_gps if gps]
-                )
+            new_assumptions.extend(self._split_assumption(ass, components))
         return (Tiling(tiling.obstructions, tiling.requirements, new_assumptions),)
 
     @staticmethod
+    def _split_assumption(assumption, components):
```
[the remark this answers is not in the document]
christianbean: Fixed.

tilings/strategies/fusion.py
```diff
@@ -1,28 +1,137 @@
...
+        # the child parameter that determine 'where the line is drawn'.
+        self.fuse_parameter = fuse_parameter
+        # sets to tell if parent assumption is one sided, or both
+        self.left_sided_parameters = frozenset(left_sided_parameters)
+        self.right_sided_parameters = frozenset(right_sided_parameters)
+        self.both_sided_parameters = frozenset(both_sided_parameters)
+        # The following assertions check that if two parent assumptions map to the
+        # sameassumption, then one of them is one sided, and the other covers both
+        # sides, OR that they are all contained fully in fuse region.
+        assert all(
```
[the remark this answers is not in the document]
christianbean: I moved it to a seperate _init_checked method. I agree this is a bit confusing, but it helps me be sure it is a case that is covered. Luckily the init method is only called the first time the constructor is made, but constructors are cached on the rules.

tilings/strategies/verification.py
christianbean: I wasn't sure how to do that in particular as assumptions are native to Tilings. Or is your point that the expected dictionary should be a method on CombinatorialClass ?

tilings/strategies/verification.py
christianbean: Agree, I'll move it there.

tilings/tiling.py
christianbean: I recall the logic of adding this was related to the fact that when we map assumptions, the assumption on the parent is not always aware if a cell is actually empty, and so when we map if forward we end up with a key error.

I remove this change and nothing is breaking in our successes, but I'm not 100% sure.

enadeau: My concern came from the fact that it is the opposite of what I did to fix to row col separation bug. It that case I took extra care to not include empty cell in maps to avoid trouble when composing maps. I guess we will see if a bug comes out.

tilings/strategies/fusion.py
```diff
@@ -1,28 +1,147 @@
...
+    """
+
+    def __init__(
+        self,
+        fuse_parameter: str,
+        extra_parameters: Dict[str, str],
+        left_sided_parameters=Iterable[str],
+        right_sided_parameters=Iterable[str],
+        both_sided_parameters=Iterable[str],
+    ):
```
enadeau, line 40: Can we add a docstring that describe what are the arguments

tilings/strategies/fusion.py
```diff
@@ -1,28 +1,147 @@
...
+        self.both_sided_parameters = frozenset(both_sided_parameters)
+
+        # In particular, the lists in reversed_extra_parameters can have size at
+        # most two (in which case  of the two variables is a subset of the other),
+        # unless, they are contained entirely within the fused region, in which
+        # case it can have up to 3, namely left, right and both. This is checked
+        # in the following function.
+        self._init_checked()
+
+        # if any two parents map to the same child, then these determine the
```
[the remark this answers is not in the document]
christianbean: Yes, made the comment clearer.

tilings/strategies/fusion.py
```diff
@@ -1,28 +1,147 @@
...
+            or (
+                len(val) == 3
+                and all(
+                    self.extra_parameters[parent_var] == self.fuse_parameter
+                    for parent_var in val
+                )
+            )
+            for val in self.reversed_extra_parameters.values()
+        )
+        for parent_var, child_var in self.extra_parameters.items():
```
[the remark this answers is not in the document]
christianbean: Ooops, you are right!

tilings/strategies/fusion.py
christianbean: Will do.

tilings/strategies/obstruction_inferral.py
```diff
@@ -34,6 +36,37 @@ def formal_step(self) -> str:
...
+                for gp in ass.gps
+                if gp.avoids(*self.gps)
+                and all(cell in av.forward_cell_map for cell in gp.pos)
+            )
+            for ass in comb_class.assumptions
+        ]
+        av_mapped_ass = [
+            TrackingAssumption(gp for gp in gps if gp.avoids(*av.obstructions))
+            for gps in av_mapped_gps
+        ]
```
enadeau, line 61: If I understand correctly this bit of code is mapping the assumption according to some cell map and then performing some cleaning based. As it seems this bit of is repeated in many the `extra_parameters` methods in should probably be factor out somewhere. Perhaps on the tracking assumption class?

enadeau: Actually, reading further I see that the entire `extra_parameters` methods is very similar on all the disjoint union based strategy. Maybe that's a track to refactoring it? I'd be happy to chat about that if you think it's a good idea.

tilings/strategies/fusion.py
```diff
@@ -1,28 +1,182 @@
...
+                                not at all. # TODO: better name?
+    """
+
+    def __init__(
+        self,
+        fuse_parameter: str,
+        extra_parameters: Dict[str, str],
+        left_sided_parameters=Iterable[str],
+        right_sided_parameters=Iterable[str],
+        both_sided_parameters=Iterable[str],
```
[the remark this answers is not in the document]
christianbean: In all its uses, we really mean covering both sides, although are in cases where we are definitely covering some part of the region to be fused. I should update this accordingly.

tilings/strategies/fusion.py
```diff
@@ -32,11 +186,260 @@ def reliance_profile(self, n: int, **parameters: int) -> RelianceProfile:
...
+            if new_params is not None:
+                assert new_params[self.fuse_parameter] == left_points + right_points
+                res += subrec(n, **new_params)
+        return res
+
+    def _determine_number_of_points_in_fuse_region(
+        self, n: int, **parameters: int
+    ) -> Iterator[Tuple[int, int]]:
+        """
+        There are two cases we use to determine the number of left and right points.
```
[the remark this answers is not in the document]
christianbean: Removed it from the top - think its better to keep it with the function.

tilings/strategies/fusion.py
```diff
@@ -32,11 +186,260 @@ def reliance_profile(self, n: int, **parameters: int) -> RelianceProfile:
...
+                max_right_points = min(
+                    max_right_points, number_points_parent_fuse_parameter
+                )
+                min_both_points = max(
+                    min_both_points, number_points_parent_fuse_parameter
+                )
+                max_both_points = min(
+                    max_both_points, number_points_parent_fuse_parameter
+                )
+            if min_left_points > max_left_points or min_right_points > max_right_points:
```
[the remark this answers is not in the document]
christianbean: added this

tilings/strategies/fusion.py
```diff
@@ -32,11 +186,260 @@ def reliance_profile(self, n: int, **parameters: int) -> RelianceProfile:
...
+                return None
+            updated_value = (
+                value + number_of_left_points
+                if parameter in self.right_sided_parameters
+                else value + number_of_right_points
+                if parameter in self.left_sided_parameters
+                else value
+            )
+            if updated_value < 0:
+                # TODO: is this reached?
```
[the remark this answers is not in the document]
christianbean: agree, removed the check

tilings/strategies/fusion.py
```diff
@@ -75,11 +478,70 @@ def decomposition_function(self, comb_class: Tiling) -> Tuple[Tiling, ...]:
...
+            right_sided_params,
+            both_sided_params,
+        )
+
+    def _fuse_parameter(self, comb_class: Tiling) -> str:
+        algo = self.fusion_algorithm(comb_class)
+        child = algo.fused_tiling()
+        fuse_assumption = algo.new_assumption()
+        idx = child.assumptions.index(fuse_assumption)
+        return "k_{}".format(idx)
```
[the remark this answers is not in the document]
christianbean: agree, fixed

tilings/strategies/fusion.py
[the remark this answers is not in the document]
christianbean: pushed the change I mentioned

tilings/strategies/obstruction_inferral.py
christianbean: I changed to use the avoiding method in TrackingAssumption.

This doesn't quite belong on the level of DisjointUnionStrategy, as all maps are specific to tilings. I am open to suggestions though. 
The process is essentially three steps:
1) map according to the bijection implied by the strategy, e.g. cell insertion is identity map, point placement is the stretching map, etc
2) map this according to the forward map of the tiling initialiser
3) remove those which contain an obstruction on the child tiling

ulfarsson: You could argue that if css is always working with combinatorial sets, then those sets always contain combinatorial objects, which can be mapped around. However, I think css can work with more general "things", where there are no underlying objects.

christianbean: Its mostly because of 2 that it shouldn't belong there. Our strategies are really always applying two strategies at one. The strategy itself, then remove empty rows and columns.

enadeau: Ok I still think some of it should be factored as there is a lot of redundant code but I think it's part a broader design discussion we should have about how we implement maps. I'll open an issue

christianbean: Ok, I fixed conflicts. Shall we delay refactoring to a future PR then, and accept this one?

enadeau: Yes I think this is ready to merge

PR #141, opened 2020-06-02
Update root pack

enadeau:
- [x] Add an option to bound the number of rules on requirement placements

- [x] Bound the number of rule on the root placement pack

### DIFF
--- a/tilings/strategies/requirement_placement.py
+++ b/tilings/strategies/requirement_placement.py
@@ -1,5 +1,6 @@
 import abc
 from collections import defaultdict
+from functools import reduce
 from itertools import chain, product
 from typing import Dict, Iterable, Iterator, List, Optional, Set, Tuple, cast
 
@@ -405,6 +406,10 @@ class RequirementPlacementFactory(AbstractRequirementPlacementFactory):
             - `permuta.misc.DIR_SOUTH`
             - `permuta.misc.DIR_EAST`
             - `permuta.misc.DIR_WEST`
+        - `max_rules_per_req`: The limit on the size of the requirements that can
+        be placed. If the product of the length of the gridded perms
+        in the requirement and the number of direction is greater than the max,
+        then no placements is performed for that req.
     """
 
     def __init__(
@@ -413,9 +418,11 @@ class RequirementPlacementFactory(AbstractRequirementPlacementFactory):
         partial: bool = False,
         ignore_parent: bool = False,
         dirs: Iterable[int] = tuple(DIRS),
+        max_rules_per_req: Optional[int] = None,
     ):
         assert all(d in DIRS for d in dirs), "Got an invalid direction"
         self.subreqs = subreqs
+        self.max_rules_per_req = max_rules_per_req
         super().__init__(partial=partial, ignore_parent=ignore_parent, dirs=dirs)
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


tilings/strategy_pack.py
```diff
@@ -276,9 +276,18 @@ def insertion_row_and_col_placements(
...
+        length: int = 3,
+        max_num_req: Optional[int] = 1,
+        max_placement_rules_per_req: Optional[int] = None,
     ) -> "TileScopePack":
-        name = "only_length_{}_root_placements".format(length)
+        if max_num_req is not None:
+            name = "only_length_{length}_{max_num_req}_reqs_root_placements"
+        else:
+            name = "only_length_{length}_root_placements"
+        placment_factory = strat.RequirementPlacementFactory(
```
jaypantone, line 288: typo

tilings/strategy_pack.py
```diff
@@ -276,9 +276,18 @@ def insertion_row_and_col_placements(
...
+        cls,
+        length: int = 3,
+        max_num_req: Optional[int] = 1,
+        max_placement_rules_per_req: Optional[int] = None,
     ) -> "TileScopePack":
-        name = "only_length_{}_root_placements".format(length)
+        if max_num_req is not None:
+            name = "only_length_{length}_{max_num_req}_reqs_root_placements"
+        else:
+            name = "only_length_{length}_root_placements"
```
jaypantone, line 287: Do we want `max_placement_rules_per_req` in the name too?

tilings/strategy_pack.py
enadeau: At least with autocomplete I'm consistent in my typos :)

tilings/strategy_pack.py
enadeau: I thought about that but it made the name too long in my opinion. We still have all the details of the pack that we ran as we save the JSON of the pack in our database

jaypantone:
I admit I never fully figured out what this pack does, so I'm not the best person to review it.

tilings/strategies/requirement_placement.py
```diff
@@ -453,6 +459,10 @@ def req_indices_and_directions_to_place(
         else:
             all_reqs = tiling.requirements
         for req in all_reqs:
+            if self.max_rules_per_req is not None:
+                num_rule = reduce(lambda prod, req: prod * len(req), req, 1)
```
christianbean, line 463: should multiply by 4 for directions

tilings/strategy_pack.py
```diff
@@ -276,9 +276,18 @@ def insertion_row_and_col_placements(
...
+        placement_factory = strat.RequirementPlacementFactory(
+            max_rules_per_req=max_placement_rules_per_req
+        )
```
christianbean, line 290: now I'm being picky, but shouldn't placement factory variable name just be skippd and it added to expansion strats directly- i don't really care, but it is the way we do it elsewhere

tilings/strategies/requirement_placement.py
enadeau: I thought about that, it depends on how we want this flag to play with the `dirs` flag. Maybe we should multiply by the number of directions.

christianbean: Good point, it should multiple len(dirs) as you suggested

tilings/strategy_pack.py
enadeau: It was ugly when I put that in the strat directly, so I moved it. Not that I really care either.

christianbean: ok, don't worry about it, lets fix the other thing and we're good to go!

tilings/strategies/requirement_placement.py
enadeau: Fixed.

PR #146, opened 2020-06-04
Improve expansion packs to handle assumptions.

enadeau:


enadeau:
We can merge this PR as is but I think it's not the right approach  to monotone tree verification. See #149

PR #150, opened 2020-06-05
Expanding locally verified objects

christianbean:
Try to expand verified classes with packs that we know will succeed in enumerating the class in terms of atoms.


tilings/strategies/verification.py
```diff
@@ -95,10 +99,71 @@ def basis(self) -> Tuple[Perm, ...]:
...
+                    )
+                    .fix_one_by_one(basis)
+                )
+        if pack is None:
+            raise InvalidOperationError(
+                "Cannot get a specification for one by one verification for "
+                f"subclass Av({basis})"
+            )
+        pack.add_initial(SplittingStrategy(ignore_parent=True), apply_first=True)
+        return pack
```
christianbean, line 166: these lines are redundant

PR #159, opened 2020-06-11
gridded perm example

christianbean:


christianbean:
decided to use ascii plot instead

PR #162, opened 2020-06-11
remove the timing on tilings methods and algorithm

enadeau:


tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -18,7 +16,6 @@ class GriddedPermReduction:
     func_calls = func_calls
     func_times = func_times
```
christianbean, line 17: We can remove these two dictionaries, as we are not timing anything.

tilings/tiling.py
```diff
@@ -94,7 +93,6 @@ class Tiling(CombinatorialClass):
     func_calls = func_calls
     func_times = func_times
```
christianbean, line 95: these can go also

christianbean: fixed

tilings/algorithms/gridded_perm_reduction.py
christianbean: fixed

enadeau:
Good catch I really forgot to remove the dict. I'll merge

PR #167, opened 2020-06-18
Add assumptions

christianbean:
- add a new `AddAssumptionStrategy` which adds an assumption to a tiling.
  In practice, when expanding a class, we actually remove an assumption to
  determine which rules to add.
- the `get_equations` method is now implemented for the strategies
  `AddAssumptionStrategy`, `SplittingStrategy`, and `FusionStrategy`.


tilings/strategies/assumption_insertion.py
```diff
@@ -0,0 +1,184 @@
...
+    SubRecs,
+    SubSamplers,
+)
+from tilings import GriddedPerm, Tiling
+from tilings.assumptions import TrackingAssumption
+
+
+class AddAssumptionConstructor(Constructor):
+    """
+    The constructor used to cound when a new variable is added.
```
enadeau, line 25: cound -> count

tilings/strategies/assumption_insertion.py
```diff
@@ -0,0 +1,184 @@
...
+        child = children[0]
+        return (
+            {
+                comb_class.get_parameter(ass): child.get_parameter(ass)
+                for ass in comb_class.assumptions
+            },
+        )
+
+    def formal_step(self) -> str:
+        return "adding the assumption '{}'".format(self.assumption)
```
enadeau, line 115: Think about the new line issue

tilings/strategies/assumption_insertion.py
```diff
@@ -0,0 +1,184 @@
...
+        objs: Tuple[Optional[GriddedPerm], ...],
+        children: Optional[Tuple[Tiling, ...]] = None,
+    ) -> GriddedPerm:
+        """
+        The forward direction of the underlying bijection used for object
+        generation and sampling.
+        """
+        if children is None:
+            children = self.decomposition_function(comb_class)
+        raise NotImplementedError
```
enadeau, line 129: Both map can be easily implemented

christianbean:
> I feel like this could conflict with PR #170 we should merge one at the time and rerun the test in between

It did in the changelog and readme. Fixed these so will now merge.

PR #171, opened 2020-06-19
Strategy pack options

christianbean:
### Added
- add partial flag to `insertion_point_placements` and
  `insertion_row_and_col_placements`

 ### Fixed
- untracked fusion packs don't add assumption strategies
- the length parameter for `all_the_strategies` is passed correctly to the
  requirement insertion strategy.

### DIFF
--- a/tilings/strategy_pack.py
+++ b/tilings/strategy_pack.py
@@ -182,7 +182,7 @@ class TileScopePack(StrategyPack):
         name = "insertion_"
         if length > 1:
             name += "length_{}_".format(length)
-        partial_str = "partial" if partial else ""
+        partial_str = "partial_" if partial else ""
         name += f"{partial_str}point_placements"
         return TileScopePack(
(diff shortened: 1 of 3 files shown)


tilings/strategy_pack.py
```diff
@@ -179,11 +176,14 @@ def point_placements(
...
     @classmethod
-    def insertion_point_placements(cls, length: int = 1) -> "TileScopePack":
+    def insertion_point_placements(
+        cls, length: int = 1, partial: bool = False
+    ) -> "TileScopePack":
         name = "insertion_"
         if length > 1:
             name += "length_{}_".format(length)
-        name += "point_placements"
+        partial_str = "partial" if partial else ""
```
enadeau, line 185: Here it should be "partial_"

christianbean: Good catch, fixed that now.

PR #172, opened 2020-06-20
Tracking component fusion

christianbean:
Included is the code required for tracking the component fusion strategy. It is currently good to go up to the enumeration. 

This untracked version of this pack was often used to find a specification for `0231_0312`. In this specification recurses back to the root. In order to recreate a tracked specification for this class I needed to add a questionable addition to the 1x1 verification strategy. 

```python
    def verified(self, tiling: Tiling) -> bool:
        return tiling.dimensions == (1, 1) and (
            frozenset(ob.patt for ob in tiling.obstructions) not in self.symmetries
            or SumComponentAssumption([GriddedPerm(Perm((0,)), ((0, 0),))])
            in tiling.assumptions
            or SkewComponentAssumption([GriddedPerm(Perm((0,)), ((0, 0),))])
            in tiling.assumptions
        )
```

Is this the best way to do this? It should probably at least check that the basis elements are indecomposable. 

We should also consider carefully when other verification strategies may or may not verify a tiling containing one of these assumptions.


tilings/assumptions.py
```diff
@@ -72,3 +80,119 @@ def __str__(self):
...
+    def __hash__(self) -> int:
+        return hash(self.gps)
+
+
+class SkewComponentAssumption(ComponentAssumption):
+    @staticmethod
+    def skew_decomposition(perm: Perm) -> List[Perm]:
+        """
+        Return the skew decomposition of the permutation.
+        TODO: this method should be on permuta.
```
enadeau, line 175: Did it https://github.com/PermutaTriangle/Permuta/pull/114

tilings/algorithms/fusion.py
```diff
@@ -405,6 +411,26 @@ def fusable(self):
         new_tiling = self._tiling.add_obstructions(self.obstructions_to_add())
         return self._tiling == new_tiling
 
+    def new_assumption(self):
+        """
+        Return the assumption that needs to counted in order to enumerate.
```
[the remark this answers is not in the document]
christianbean: fixed

tilings/assumptions.py
```diff
@@ -48,18 +51,23 @@ def to_jsonable(self) -> dict:
         return {"gps": [gp.to_jsonable() for gp in self.gps]}
 
     @classmethod
-    def from_dict(cls, d: dict) -> "TrackingAssumption":
+    def from_dict(cls, d: dict) -> Union["TrackingAssumption", "ComponentAssumption"]:
```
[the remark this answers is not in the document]
christianbean: fixed

tilings/assumptions.py
```diff
@@ -72,3 +80,119 @@ def __str__(self):
         return "can count occurrences of\n{}".format(
             "\n".join(str(gp) for gp in self.gps)
         )
+
+
+class ComponentAssumption(TrackingAssumption):
+    """
+    An assumption used to keep track of the number of omponents in a
```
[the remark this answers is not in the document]
christianbean: fixed

tilings/assumptions.py
```diff
@@ -72,3 +80,119 @@ def __str__(self):
...
+    most strategies work without change.
+    """
+
+    def __init__(self, gps: Iterable[GriddedPerm]):
+        super().__init__(gps)
+        assert all(len(gp) == 1 for gp in self.gps)
+        self.cells = frozenset(gp.pos[0] for gp in self.gps)
+
+    @abc.abstractproperty
+    def decomposition(self):
```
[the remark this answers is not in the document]
christianbean: It is used to determine the number of components in a specific gridded perm.

tilings/assumptions.py
```diff
@@ -72,3 +80,119 @@ def __str__(self):
...
+        pass
+
+    def get_value(self, gp: GriddedPerm) -> int:
+        """
+        Return the number of occurrences of each of the gridded perms being track in
```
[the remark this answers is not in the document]
christianbean: fixed

tilings/assumptions.py
```diff
@@ -72,3 +80,119 @@ def __str__(self):
...
+        the gridded perm gp.
+        """
+        subgp = gp.get_gridded_perm_in_cells(self.cells)
+        return len(self.decomposition(subgp.patt))
+
+    def to_jsonable(self) -> dict:
```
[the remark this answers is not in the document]
christianbean: It is much cleaner.

tilings/assumptions.py
```diff
@@ -72,3 +80,119 @@ def __str__(self):
...
+        return {"cells": tuple(self.cells)}
+
+    @classmethod
+    def from_dict(
+        cls, d: dict
+    ) -> Union["SumComponentAssumption", "SkewComponentAssumption"]:
```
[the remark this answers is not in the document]
christianbean: the method is gone now anyway!

tilings/algorithms/fusion.py
```diff
@@ -405,6 +411,26 @@ def fusable(self):
...
+        """
+        gps = (
+            GriddedPerm.single_cell(Perm((0,)), cell)
+            for cell in self._tiling.active_cells
+            if (self._fuse_row and cell[1] == self._row_idx)
+            or (not self._fuse_row and cell[0] == self._col_idx)
+        )
```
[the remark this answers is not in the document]
christianbean: I believe so, fixed.

tilings/assumptions.py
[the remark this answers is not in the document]
christianbean: I made an abstract method, and gave it a doc string.

tilings/assumptions.py
christianbean: I changed to use this, but of course now it won't be able to pass travis tests as that points to master I think.

enadeau: Yep we should release a new version of permuta

tilings/assumptions.py
```diff
@@ -45,21 +49,36 @@ def get_value(self, gp: GriddedPerm) -> int:
...
+        }
 
     @classmethod
     def from_dict(cls, d: dict) -> "TrackingAssumption":
+        """Return the assumption from the json dict representation."""
+        module = import_module(d["class_module"])
+        AssClass: Type["TrackingAssumption"] = getattr(module, d["assumption"])
+        assert issubclass(
+            AssClass, TrackingAssumption
+        ), "Not a valid CombinatorialClass"
```
enadeau, line 67: The error message is not quite right

tilings/assumptions.py
enadeau: You should also remove the two static methods

christianbean: Thanks!

tilings/assumptions.py
christianbean: fixed

tilings/assumptions.py
enadeau: You forgot to remove the static methods :)

christianbean: good spot!

tilings/strategies/verification.py
```diff
@@ -184,9 +185,12 @@ def pack(tiling: Tiling) -> StrategyPack:
         )
 
     def verified(self, tiling: Tiling) -> bool:
-        return (
-            tiling.dimensions == (1, 1)
-            and frozenset(ob.patt for ob in tiling.obstructions) not in self.symmetries
+        return tiling.dimensions == (1, 1) and (
```
enadeau, line 188: I'm kind of surprised that we verify the root if it has an assumption on it. I feels circular

christianbean: See my initial comment on the PR.

enadeau:
We should discuss the verification strategy in meeting.

Also, I'm pretty sure this is breaking backward compatibility for load specification json. This might not be a big concern at the moment, but we should be careful about that moving forward.

christianbean:
> We should discuss the verification strategy in meeting.
> 
> Also, I'm pretty sure this is breaking backward compatibility for load specification json. This might not be a big concern at the moment, but we should be careful about that moving forward.

it does, but only those which have used assumptions.

PR #173, opened 2020-06-22
Interleaving factors

christianbean:
implemented tracking and counting for interleaving factors


tilings/strategy_pack.py
```diff
@@ -56,6 +64,43 @@ def make_fusion(
...
         return pack
 
+    def make_interleaving(self, tracked=True) -> "TileScopePack":
+        """
+        Return a new pack where the factor strategy is replaced with an
+        interleaving factor strategy.
+        """
+
+        def replace_list(strats):
+            """Return a new list with the replaced 1x1 strat."""
```
enadeau, line 74: I think the docstring is wrong, smell like copy-pasting ;)

tilings/tiling.py
```diff
@@ -945,7 +951,9 @@ def component_fusion(self, row=None, col=None):
         """
         return self._fusion(row, col, ComponentFusion)
 
-    def sub_tiling(self, cells: Iterable[Cell], factors: bool = False) -> "Tiling":
+    def sub_tiling(
+        self, cells: Iterable[Cell], factors: bool = False, add_assumptions=tuple()
```
enadeau, line 955: Can you add the typing for the `add_assumptions` argument?

tilings/strategies/assumption_insertion.py
```diff
@@ -181,3 +198,51 @@ def to_jsonable(self) -> dict:
...
+        """
+        cols, rows = interleaving_rows_and_cols(components)
+        assumptions = chain.from_iterable(
+            assumptions_to_add(cells, cols, rows) for cells in components
+        )
+        if assumptions:
+            strategy = AddAssumptionsStrategy(assumptions, workable=True)
+            yield strategy(comb_class)
+
+    # TODO: unions? monotone?
```
enadeau, line 222: What about this todo? I think it's doing unions but it's not doing monotone

tilings/strategies/factor.py
```diff
@@ -249,22 +409,33 @@ def __init__(
         self.unions = unions
         self.ignore_parent = ignore_parent
         self.workable = workable
+        self.tracked = tracked and self.factor_class in (
+            FactorWithInterleavingStrategy,
+            FactorWithMonotoneInterleaving,
+        )
 
     def __call__(self, comb_class: Tiling, **kwargs) -> Iterator[Strategy]:
+        # TODO: if tracked, ensure that assumptions are present before factoring
```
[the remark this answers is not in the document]
christianbean: no longer relevant

tilings/strategies/factor.py
```diff
@@ -154,11 +242,24 @@ def get_equation(lhs_func: Function, rhs_funcs: Tuple[Function, ...]) -> Eq:
...
+            [
+                self.count_interleavings([parameters[k] for k in int_parameters])
+                for int_parameters in self.interleaving_parameters
+            ],
+            1,
+        )
+        return multiplier * super().get_recurrence(subrecs, n, **parameters)
+
+    @staticmethod
+    def count_interleavings(values: List[int]):
```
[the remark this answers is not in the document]
christianbean: I thought it made it more readable, but you can remove it if you don't think it helps.

tilings/strategy_pack.py
christianbean: Yes, a copy paste error :(

tilings/tiling.py
christianbean: Should be a Tuple[TrackingAssumptions]

tilings/strategies/assumption_insertion.py
christianbean: Are you adding this when you add interleaving splitting?

tilings/strategies/factor.py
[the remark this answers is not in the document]
enadeau: Any way is fine for me

tilings/strategies/assumption_insertion.py
```diff
@@ -181,3 +198,51 @@ def to_jsonable(self) -> dict:
     @classmethod
     def from_dict(cls, d: dict) -> "AddAssumptionFactory":
         return cls()
+
+
+class AddInterleavingAssumptionFactory(StrategyFactory[Tiling]):
+    def __init__(self, unions: bool = True):
```
enadeau, line 204: I think the default for unions should be false

christianbean:
> I left a few comments, we should also think about updating the CHANGELOG.
> 
> Also, I'm not sure about the constructor for the untracked version. Could we just feed dummy parameters to the constructor like we do with untracked fusion? Or maybe it's the way we do it with fusion that is not optimal
> 
> I'll now update the splitting strategy to handle interleaving

I think the way we do it with Fusion is less optimal, because it actually just throws unexplained errors.

tilings/strategies/factor.py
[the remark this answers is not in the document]
christianbean: Added a comment and removed the redundant function.

tilings/strategy_pack.py
christianbean: fixed

tilings/strategies/assumption_insertion.py
christianbean: sure

tilings/strategies/assumption_insertion.py
christianbean: unions are done, monotone not. Left this as a todo- currently there is no purpose in using monotone. When it can do something more than interleaving in terms of enumeration we can come back and think about this. At the very least, using unions will certainly suffice when using monotone interleaving

PR #179, opened 2020-06-25
allow fusing rows/cols which are positive on one or both sides

christianbean:
### Added 
- Allow fusing rows and columns which are positive on either or both sides.


tilings/strategies/requirement_placement.py
```diff
@@ -623,7 +623,7 @@ def req_indices_and_directions_to_place(
                 indices = tuple(0 for _ in gps)
                 yield tuple(gps), indices, direction
         if self.place_row:
-            row_dirs = tuple(d for d in self.dirs if d in (DIR_NORTH, DIR_SOUTH))
+            row_dirs = tuple(d for d in self.dirs if d in (DIR_SOUTH,))
```
enadeau: This was probably committed by accident

christianbean: it was just fixed it - very eagle eyed!

PR #186, opened 2020-06-25
Fix 1x1 packs

christianbean:


tilings/strategies/verification.py
```diff
@@ -161,10 +161,7 @@ def pack(tiling: Tiling) -> StrategyPack:
             and len(tiling.requirements[0]) == 1
             and len(tiling.requirements[0][0]) == 1
         ):
-            if not tiling.requirements and basis in (
-                [Perm((0, 1, 2))],
-                [Perm((2, 1, 0))],
-            ):
+            if basis in ([Perm((0, 1, 2))], [Perm((2, 1, 0))]):
```
jaypantone, line 164: So this will only happen if it's Av(123) with no requirements, or Av(123) with a point requirement, but not with any other requirements?

christianbean: exactly, the if statement just prior makes sure it is either no reqs, or just a single point req.

jaypantone: Makes sense.

PR #188, opened 2020-06-25
Fix local default setting

christianbean:
### Changed
- the default in `LocalVerificationStrategy` is now `no_factors=False`.


christianbean:
this will be merged with PR #191 with a test included there.

PR #194, opened 2020-06-26
Symmetry fix

jaypantone:
Fixes a bug that mean we were never computing all 8 symmetries, just 4 (the rotations). Also fixes a two minor typos I happened to spot.


jaypantone:
> can you add something to the changelog for posterity?

Whoops! Yes.

Issue #147, opened 2020-06-04
Review all places that Tiling constructor is used to check whether init work can be reduced.


jaypantone:
For example
```python
    def can_add_obstruction(obstruction: GriddedPerm, tiling: "Tiling") -> bool:
        """Return true if `obstruction` can be added to `tiling`."""
        new_reqs = tiling.__class__.sort_requirements(
            tiling.requirements + tuple((gp,) for gp in obstruction.factors())
        )
        containing_tiling = tiling.__class__(
            tiling.obstructions,
            new_reqs,
            already_minimized_obs=True,
            derive_empty=False,
            sorted_input=True,
        )
        return containing_tiling.is_empty()
```
instead of
```python
def can_add_obstruction(obstruction: GriddedPerm, tiling: "Tiling") -> bool:
        """Return true if `obstruction` can be added to `tiling`."""
        return tiling.add_requirement(obstruction.patt, obstruction.pos).is_empty()
```

jaypantone:
Note also that sometimes simplification is not necessary, e.g., for inferral, is_empty works just fine on unsimplified Tilings, and I suspect the slight slowdown it causes is less than the slowdown of simplifying.

jaypantone:
I was going to close this because it's so old, but actually the example I gave above still looks like it can be easily sped up?

christianbean:
This is something we need to keep in mind when creating future calls to the init.

PR #195, opened 2020-06-26
Sanity check rules

christianbean:
This relies on the PR sanity check multiple variables #147 on CSS. Will need to wait on a release therefore.

### DIFF
--- a/tilings/tiling.py
+++ b/tilings/tiling.py
@@ -1159,7 +1159,10 @@ class Tiling(CombinatorialClass):
             len(gp) > 1
             for gp in chain.from_iterable(ass.gps for ass in self.assumptions)
         ):
-            raise NotImplementedError
+            raise NotImplementedError(
+                "possible parameters only implemented for assumptions with "
+                "size one gridded perms"
+            )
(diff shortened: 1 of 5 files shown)


tilings/tiling.py
```diff
@@ -1154,6 +1154,11 @@ def partial_place_col(self, idx: int, direction: int) -> Tuple["Tiling", ...]:
     def extra_parameters(self) -> Tuple[str, ...]:
         return tuple("k_{}".format(i) for i in range(len(self._assumptions)))
 
+    def possible_parameters(self, n: int) -> Iterator[Dict[str, int]]:
+        parameters = [self.get_parameter(ass) for ass in self.assumptions]
+        for values in product(*[range(n + 1) for _ in parameters]):
```
enadeau, line 1159: For this function to be correct with assumptions of greater length we need to be more clever here. I think at the moment it would be ok to simply raise an `NotImplementedError` if any assumption contains gridded perms of length >2

christianbean: Thats fair, I'll add it.

PR #204, opened 2020-06-29
Update to CSS 1.2.0

christianbean:


enadeau:
It looks good we're just missing Travis test

PR #226, opened 2020-07-08
introduces isolation levels to fusion

jaypantone:


tilings/algorithms/fusion.py
```diff
@@ -273,6 +292,39 @@ def new_assumption(self):
...
+            fusing_cells = [
+                (self._col_idx, i) for i in range(self._tiling.dimensions[1])
+            ] + [(self._col_idx + 1, i) for i in range(self._tiling.dimensions[1])]
+        num_fusing_assumptions = len(
+            [
+                assumption
+                for assumption in self._tiling.assumptions
+                if any(cell in gp.pos for gp in assumption.gps for cell in fusing_cells)
+            ]
+        )
```
christianbean, line 316: I would create a method `_num_fusing_assumptions` then you will only make this if needed (i.e., in 'isolated' you first check if the number of assumptions are ok)

jaypantone: Done.

PR #228, opened 2020-07-10
added html representation

annijaapine:


tilings/tiling.py
```diff
@@ -1516,6 +1516,64 @@ def get_genf(self, *args, **kwargs) -> sympy.Expr:
             "We were unable to enumerate this tiling:\n" + str(self)
         )
 
+    def to_html_representation(self):
```
enadeau, line 1519: Can you add the output type as a string with `-> str`

Issue #144, opened 2020-06-04
Pylint issue on tilings


enadeau:
I think it's time to turn back on some of the pylint check on the tiling file

jaypantone:
After the tiling-caching PR is approved, the following pylint ignores are all present and necessary to pass tests:
```
# pylint: disable=too-many-locals
# pylint: disable=too-many-lines
# pylint: disable=too-many-branches
# pylint: disable=arguments-differ
# pylint: disable=too-many-statements
# pylint: disable=import-outside-toplevel
```

christianbean:
Some, e.g. `pylint: disable=import-outside-toplevel` should be made local to the issue in get genf

PR #230, opened 2020-07-13
pylint fixes in tiling.py

christianbean:
Removes some pylint errors, and moves failures to local to the function which have them. 

This is in response to issue #144

### DIFF
--- a/tilings/tiling.py
+++ b/tilings/tiling.py
@@ -1,9 +1,4 @@
-# pylint: disable=too-many-locals
 # pylint: disable=too-many-lines
-# pylint: disable=too-many-branches
-# pylint: disable=arguments-differ
-# pylint: disable=too-many-statements
-# pylint: disable=import-outside-toplevel
 import json
 from array import array
 from collections import Counter, defaultdict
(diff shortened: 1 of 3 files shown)


tilings/tiling.py
```diff
@@ -395,16 +390,10 @@ def split_16bit(n) -> Tuple[int, int]:
...
 
     @classmethod
-    def from_bytes(
-        cls,
-        arrbytes: bytes,
-        remove_empty_rows_and_cols=False,
-        derive_empty=False,
-        simplify=False,
-        sorted_input=True,
-    ) -> "Tiling":
```
enadeau, line 405: I don't like removing the arguments. Explicit is better than implicit. It makes the code less readable and you need to actually read the code to know how to use it. Putting the argument in the docstring would help but still having them is the best.

Maybe better is to realize that we don't need those arguments since we can assume that when we call `to_bytes` the tiling has been properly initialized.

christianbean: I agree with your solution, I at least never used another setting.

Issue #154, opened 2020-06-09
Counting bug


jaypantone:
To recreate:
```python
from permuta import Perm
from tilings import *
from tilings.strategies import *
from tilings.tilescope import TileScope, TileScopePack
basis = "0123_0213"
pack = TileScopePack.row_and_col_placements(row_only=True, partial=True).add_initial(
        PatternPlacementFactory(point_only=True)
    ).make_fusion(tracked=False)
css = TileScope(basis, pack, ruledb="forget")
spec = css.auto_search(status_update=10, max_time=120)
print([spec.count_objects_of_size(i) for i in range(11)])
```

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-54fa667dd7c7> in <module>
      9 css = TileScope(basis, pack, ruledb="forget")
     10 spec = css.auto_search(status_update=10, max_time=120)
---> 11 print([spec.count_objects_of_size(i) for i in range(11)])

<ipython-input-1-54fa667dd7c7> in <listcomp>(.0)
      9 css = TileScope(basis, pack, ruledb="forget")
     10 spec = css.auto_search(status_update=10, max_time=120)
---> 11 print([spec.count_objects_of_size(i) for i in range(11)])

~/Dropbox/Research/Active/2017-44-ATRAP/repos/comb_spec_searcher/comb_spec_searcher/specification.py in count_objects_of_size(self, n, **parameters)
    276         limit = n * self.number_of_rules()
    277         with RecursionLimit(limit):
--> 278             return self.root_rule.count_objects_of_size(n, **parameters)
    279
    280     def generate_objects_of_size(

~/Dropbox/Research/Active/2017-44-ATRAP/repos/comb_spec_searcher/comb_spec_searcher/strategies/rule.py in count_objects_of_size(self, n, **parameters)
    376                 self.subrecs is not None
    377             ), "you must call the set_subrecs function first"
--> 378             res = self.constructor.get_recurrence(self.subrecs, n, **parameters)
    379             self.count_cache[key] = res
    380         #     # THE FOLLOWING CODE SNIPPET IS FOR DEBUGGING PURPOSES

~/Dropbox/Research/Active/2017-44-ATRAP/repos/comb_spec_searcher/comb_spec_searcher/strategies/constructor.py in get_recurrence(self, subrecs, n, **parameters)
    327     def get_recurrence(self, subrecs: SubRecs, n: int, **parameters: int) -> int:
[...]

PR #231, opened 2020-07-13
Update to CSS 2.0.0

christianbean:
### Changed
- the `get_eq_symbol` and `get_op_symbol` are moved to `Strategy` rather than
  `Constructor`

 ### Fixed
- untracked constructors raise `NotImplementedError`


christianbean:
This "fixes" (/implements the not implemented error promised in) issue #154

christianbean:
The tests fail, because the string method of spec is failing! We need a release of CSS to pass :(

PR #235, opened 2020-07-17
Int factor inferrable

enadeau:


enadeau:
Those change have been integrated in #234  so closing

PR #238, opened 2020-07-17
Normal rule when no interleaving

enadeau:


tilings/strategies/factor.py
```diff
@@ -224,7 +224,7 @@ def get_equation(lhs_func: Function, rhs_funcs: Tuple[Function, ...]) -> Eq:
         raise NotImplementedError
 
     def get_recurrence(self, subrecs: SubRecs, n: int, **parameters: int) -> int:
-        # multinomial counts the number of ways to interleave the values k1, ..., kn.
+        # multinomial counts the number of ways to interleave the values k, ..., kn.
```
christianbean, line 227: I think you changed this by mistake

enadeau: Whattt! I'll fix that

jaypantone: Why did k1 turn into k here?

PR #240, opened 2020-07-17
Positive fusion eqs

jaypantone:
Enables fusion equations when one or both of the fusing cells is positive.


jaypantone:
Needs tests.

PR #242, opened 2020-07-20
Remove length

### LINKED ISSUE #241: insertion_point_placements(2) does not work / is not implemented
I think this is because it has the strategy
```python
strat.CellInsertionFactory(
    maxreqlen=length, ignore_parent=True, one_cell_only=False
),
```
which has the lines
```python
if self.one_cell_only:
   assert self.maxreqlen == 1 and self.ignore_parent
```

Not sure what our goal is here.

christianbean:
removed the length argument to the `insertion_point_placements` pack which was not implemented, and thus raising an error.

fixes #241

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   `ComponentAssumption`
 - a tiling factors if a `ComponentAssumption` if the components of the region
   split into the factors
-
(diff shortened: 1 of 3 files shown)


christianbean:
> Looks good. Is your thinking that insertion_point_placements(2) was not a pack accomplishing much anyway?

yes, as it will only work on tilings that must contain every possible size 2 local gridded perm avoiding the obstructions. That is not a smart strategy pack.

PR #243, opened 2020-07-20
update to permuta 2.0.1

christianbean:


christianbean:
> As the change are not affecting the user at all, I'm not sure it should appear in the CHANGELOG

I agree with the logic, we can remove the changelog entry.

PR #247, opened 2020-07-23
Subclass verification

jaypantone:


tilings/strategies/experimental_verification.py
```diff
@@ -0,0 +1,61 @@
+"""
+This file contains "experimental" verification packs, which are those that verify
```
enadeau, line 2: *verification strategies

tilings/strategies/experimental_verification.py
```diff
@@ -0,0 +1,61 @@
...
+    def formal_step() -> str:
+        return "tiling is a subclass"
+
+    def __str__(self) -> str:
+        return "subclass verification"
+
+    def to_jsonable(self) -> dict:
+        d: dict = super().to_jsonable()
+        d["perms_to_check"] = (
+            frozenset(self.perms_to_check) if self.perms_to_check is not None else None
```
enadeau, line 54: Did you test the json mehtod? I'm not sure frozenset are json serializable. Perhaps you can add a test to the `test_json_encoding.py` in `/tests/strategies`

tilings/strategies/experimental_verification.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+        return d
+
+    @classmethod
+    def from_dict(cls, d: dict) -> "SubclassVerificationStrategy":
+        perms_to_check = d.pop("perms_to_check")
```
enadeau, line 60: Also think you'll need to build perm object from the json here if you want this to work properly. You can probably copy from OneByOne verification

tilings/strategy_pack.py
```diff
@@ -38,6 +41,47 @@ def replace_list(strats):
...
+           has length strictly smaller than the maximum length cell basis element.
+        """
+
+        def alter_list(strats) -> None:
+            """
+            Find subclass verification and alter its perms_to_check variable.
+            """
+            for strategy in strats:
+                if isinstance(strategy, strat.SubclassVerificationStrategy):
+                    if strategy.perms_to_check is None:
```
enadeau, line 60: We should be careful to create a new strategy and replacing the other instead of editing this strategy in place. Editing in place will create problem if we use the same pack in multiple searcher at the same time

tilings/strategies/experimental_verification.py
```diff
@@ -0,0 +1,61 @@
...
+        super().__init__(ignore_parent=ignore_parent)
+
+    def verified(self, tiling: Tiling) -> bool:
+        assert self.perms_to_check is not None, "perms_to_check was never set"
+        algo = SubclassVerificationAlgorithm(tiling, self.perms_to_check)
+        return algo.verified()
+
+    @staticmethod
+    def formal_step() -> str:
+        return "tiling is a subclass"
```
christianbean, line 46: it might be useful to see which subclass. You can have the algo save perms left as a property perhaps?

tilings/strategy_pack.py
```diff
@@ -38,6 +41,47 @@ def replace_list(strats):
...
+                        perm_gen = Perm.up_to_length(max_length)
+                        for perm in perm_gen:
+                            if len(perm) == 0:
+                                continue
+                            if any(perm.avoids(*basis) for basis in cell_bases):
+                                strategy.perms_to_check.add(perm)
+                    logger.info(
+                        "SubclassVerification set up to check the subclasses: Av(%s)",
+                        "), Av(".join(map(str, strategy.perms_to_check)),
+                    )
```
christianbean, line 77: If you send a 1x1, e.g. Av(1234) this prints 
Av(0), Av(01), Av(10), Av(012), Av(021), Av(102), Av(120), Av(201), Av(210)
It would be cleaner to say: proper subclasses of Av(0123).

tilings/algorithms/subclass_verification.py
```diff
@@ -0,0 +1,91 @@
...
+            return False
+
+        perms_to_check = self.quick_pare()
+        if len(perms_to_check) == 0:
+            return False
+
+        max_len_of_perms_to_check = max(map(len, perms_to_check))
+        max_length = (
+            self.tiling.maximum_length_of_minimum_gridded_perm()
+            + max_len_of_perms_to_check
```
christianbean, line 74: this can be -1, as you have to use at least one point of a minimum gridded perm, right?

tilings/algorithms/subclass_verification.py
```diff
@@ -0,0 +1,91 @@
...
+        )
+
+        GP = GriddedPermsOnTiling(self.tiling).gridded_perms(
+            max_length, place_at_most=max_len_of_perms_to_check
```
christianbean, line 78: this can also be `place_at_most=max_len_of_perms_to_check - 1`

tilings/algorithms/subclass_verification.py
```diff
@@ -0,0 +1,91 @@
...
+                    perms_leftover.add(perm)
+            return perms_leftover
+
+        return self.perms_to_check
+
+    def verified(self) -> bool:
+        if len(self.tiling.point_cells) > 0:
+            return False
+        if Factor(self.tiling).factorable():
+            return False
```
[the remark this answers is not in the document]
jaypantone: Yes. Now I'm realizing that the point_cell check will also prevent the algorithm from running when a point has been partially placed in a way that won't factor. Which I probably don't want.

tilings/strategies/experimental_verification.py
jaypantone: I don't exactly see how to do this. There's a single `SubclassVerificationStrategy` object doing all of the verification, right?

christianbean: This is a minor thing, so if it is too much effort we can skip. 

My understanding is that at the end of the verified method on `SubclassVerificationAlgorithm`, we know that it is a subclass of `Av(perms_left)`. 

So I see two options:
1) store the perms left as an attribute on the algorithm
2) have a method on the algorithm called `_perms_avoided` which returns `perms_left`. The verification is then simply `return bool(self._perms_avoided())`, and the string we want is f`is subclass of {Av(algo._perms_avoided())}'.

tilings/algorithms/subclass_verification.py
```diff
@@ -58,15 +59,24 @@ def quick_pare(self) -> Set[Perm]:
...
+    def compute_subclasses(self) -> None:
+
+        self._subclasses = tuple()
+
         # It is a waste of time to check a factorable tiling, since we will check its
         # children eventually.
-        if len(self.tiling.point_cells) > 0 or Factor(self.tiling).factorable():
-            return False
+        if Factor(self.tiling).factorable():
+            return
```
christianbean, line 75: Should this check happen on the strategy. We might want to add a `Tiling.subclass_verified` for example.

tilings/strategies/experimental_verification.py
```diff
@@ -20,44 +20,86 @@
...
 
-    def verified(self, tiling: Tiling) -> bool:
+    def __call__(
+        self, comb_class: Tiling, **kwargs
+    ) -> Iterator[SubclassVerificationStrategy]:
         assert self.perms_to_check is not None, "perms_to_check was never set"
-        algo = SubclassVerificationAlgorithm(tiling, self.perms_to_check)
-        return algo.verified()
+        algo = SubclassVerificationAlgorithm(comb_class, self.perms_to_check)
+        if len(algo.subclasses) > 0:
```
christianbean, line 83: this should just be `if algo.subclasses`

tilings/algorithms/subclass_verification.py
```diff
@@ -0,0 +1,100 @@
...
+            self.compute_subclasses()
+        return cast(Tuple[Perm, ...], self._subclasses)
+
+    def compute_subclasses(self) -> None:
+
+        self._subclasses = tuple()
+
+        # It is a waste of time to check a factorable tiling, since we will check its
+        # children eventually.
+        if Factor(self.tiling).factorable():
```
enadeau, line 74: I'm think it would be better to check that in the Factory instead. This way we can still use the algorithm on a factorable tiling if we want

tilings/strategies/experimental_verification.py
```diff
@@ -0,0 +1,122 @@
...
+
+from typing import Iterable, Iterator, Optional
+
+from comb_spec_searcher import StrategyFactory, VerificationStrategy
+from permuta import Basis, Perm
+from tilings import GriddedPerm, Tiling
+from tilings.algorithms import SubclassVerificationAlgorithm
+
+__all__ = [
+    "SubclassVerificationStrategy",
```
enadeau, line 15: This should be the Factory now I think

tilings/strategies/experimental_verification.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def __init__(
+        self, subclass_basis=Iterable[Perm], ignore_parent: bool = True,
+    ):
+        self.subclass_basis = tuple(sorted(subclass_basis))
+        super().__init__(ignore_parent=ignore_parent)
+
+    @staticmethod
+    def verified(tiling: Tiling) -> bool:
+        return True
```
enadeau, line 34: I'd take @christianbean opinion but this feels weird to me. Should we check that the strategy applies?

tilings/strategies/experimental_verification.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def formal_step(self) -> str:
+        return "tiling is contained in the subclass Av({})".format(
```
enadeau, line 37: I think the string would be cleaner with this way
`tiling is contained in the subclass {}".format(Av(self.subclass_basis))`

tilings/strategies/experimental_verification.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def __init__(
+        self, subclass_basis=Iterable[Perm], ignore_parent: bool = True,
+    ):
+        self.subclass_basis = tuple(sorted(subclass_basis))
+        super().__init__(ignore_parent=ignore_parent)
+
+    def verified(self, tiling: Tiling) -> bool:
+        algo = SubclassVerificationAlgorithm(tiling, set(self.subclass_basis))
+        return algo.subclasses == self.subclass_basis
```
christianbean, line 35: thought we said it should be algo.is_verified()

christianbean: I realise you need all patterns to be in perm left to guarantee you avoid all of them. Ignore me!

PR #250, opened 2020-07-30
Strat pack changes

jaypantone:
- combines the expansion strategies of most packs into a single expansion set
- adds the point_and_row_and_col_placements strategy that has proved useful in the last 2x4 and several forests

### DIFF
--- a/tilings/strategy_pack.py
+++ b/tilings/strategy_pack.py
@@ -178,8 +178,10 @@ class TileScopePack(StrategyPack):
                 strat.ObstructionTransitivityFactory(),
             ],
             expansion_strats=[
-                [strat.RequirementInsertionFactory(maxreqlen=length)],
-                [strat.AllPlacementsFactory()],
+                [
+                    strat.RequirementInsertionFactory(maxreqlen=length),
+                    strat.AllPlacementsFactory(),
+                ],
             ],
             name="all_the_strategies",
         )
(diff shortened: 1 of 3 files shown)


tilings/strategy_pack.py
```diff
@@ -411,8 +415,55 @@ def requirement_placements(
...
+                strat.LocallyFactorableVerificationStrategy(),
+            ],
+            inferral_strats=[
+                strat.RowColumnSeparationStrategy(),
+                strat.ObstructionTransitivityFactory(),
+            ],
+            expansion_strats=[
+                [
+                    strat.CellInsertionFactory(maxreqlen=length),
+                    strat.PatternPlacementFactory(),
```
christianbean, line 464: Should the partial flag be added to the pattern placement strategy?

jaypantone: I was trying to mimic `point_placements`, which doesn't have it there. Is that an error?

I was also looking at how the partial packs are handled in the hammer, e.g., 
```python
TileScopePack.row_and_col_placements(partial=True).add_initial(
            PatternPlacementFactory(point_only=True)
        ),
```
so I'm not sure if
```python
TileScopePack.point_and_row_and_col_placements(partial=True),
```
is good enough? Or do I need the `point_only=True` flag passed somewhere?

PR #251, opened 2020-07-30
new default behavior for req insertion

jaypantone:


jaypantone:
@christianbean Good catch. I had forgotten to add new to_jsonable and from_dict methods, which had just been inheriting from the superclass.

PR #257, opened 2020-08-12
Generating function for 1x1 with assumptions

enadeau:
I would be nice to get be able to get the generation function for 1x1 with assumption. I've started doing it but the problem is a bit more involved than I expected so I have the test but still need to do find the fix.


enadeau:
The bug is now fix but it was a bug on css. The test will pass when PermutaTriangle/comb_spec_searcher#213 makes it to the release.

PR #261, opened 2020-09-09
Gridded perm from iterable

enadeau:


ulfarsson (approved):
Looks good. I like how this makes the code more readable

PR #274, opened 2020-10-07
Tmap

JonSteinn:


tilings/tiling.py
```diff
@@ -345,6 +346,128 @@ def _clean_assumptions(self) -> None:
...
+        by all.
+        """
+        gps = set(gps)
+        c, r, longest = cls._get_dimensions(gps)
+        max_len = longest if max_len == -1 else min(max_len, longest)
+        obs = cls._search(deque([GriddedPerm()]), gps, c, r, max_len)
+        return cls(obstructions=obs, requirements=(), assumptions=())
+
+    @staticmethod
+    def _extend_gp(gp: GriddedPerm, c: int, r: int) -> Iterator[GriddedPerm]:
```
enadeau, line 368: This should probably be turned into a method of `GriddedPerm`

christianbean: A similar function `GriddedPerm.insert_specific_point` might simplify this

JonSteinn: Fixed

JonSteinn:
> This is a bit to much code to add to the tiling file. Can we only keep the `guess_from_gridded_perms` method and move all the logic to an algorithm file?

Done

enadeau (approved):

PR #276, opened 2020-10-25
Slide

JonSteinn:


enadeau:
We will close that and integrate that with the future slide algorithm/strategy

PR #277, opened 2020-10-26
Generating and sampling permutations from specifications that use the fusion strategy

christianbean:
Allow for generation and sampling of specifications with assumptions and using the fusion strategy. 

This relies on a release of CSS after the PR that updates sampling and generation to allow for multiple parameters.


enadeau (approved):
I approve the changes @christianbean will push

PR #284, opened 2020-11-19
reformatting param map

enadeau:
This goes with the similar PR on css.

Also included a small refactoring in tests.


christianbean (approved):

enadeau:
Great we will be able to merge that after the next css release

PR #285, opened 2020-11-19
Slide Strategy

JonSteinn:


tilings/algorithms/__init__.py
```diff
@@ -35,4 +43,10 @@
     "RowColSeparation",
     "SubclassVerificationAlgorithm",
     "guess_obstructions",
+    "get_col_info",
+    "slide_column",
+    "slidable_pairs",
+    "slide_assumption",
+    "gp_slide",
+    "gp_slide_inverse",
```
christianbean, line 51: Most of these functions are too specific to the sliding stuff. I think we should think about creating a `Sliding` class for containing the algorithm and these methods (more akin to the way the other algorithms are done).

tilings/strategies/sliding.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+def _tiling_rotate90_and_reverse(tiling: Tiling) -> Tiling:
+    t: Tiling = tiling.rotate90().reverse()
+    return t
+
+
+def _tiling_rotate90_and_reverse_inverse(tiling: Tiling) -> Tiling:
+    t: Tiling = tiling.reverse().rotate270()
+    return t
```
christianbean, line 44: Do these really need functions?

tilings/strategies/sliding.py
```diff
@@ -0,0 +1,276 @@
...
+    return _tmp_func
+
+
+def _gp_rotate90_and_reverse_inverse(r: int) -> Callable[[GriddedPerm], GriddedPerm]:
+    def _tmp_func(gp: GriddedPerm) -> GriddedPerm:
+        return GriddedPerm(
+            gp.patt.reverse().rotate(-1), ((0, r - x - 1) for x, _ in gp.pos)
+        )
+
+    return _tmp_func
```
christianbean, line 87: I understand why you have given these their own methods. The methods on `GriddedPerm` maybe should take into account the context in which the symmetry is taken in a simple way. This is probably an issue for a future PR.

tilings/strategies/sliding.py
christianbean: Even the case where you apply two symmetries could be a single application of one symmetry - i.e. the anti-diagonal or complement symmetries (I think).

tilings/strategies/sliding.py
```diff
@@ -0,0 +1,276 @@
...
+            _gp_rotate90_and_reverse(r),
+            _gp_rotate90_and_reverse_inverse(r),
+        )
+
+
+class SlidingStrategy(DisjointUnionStrategy[Tiling, GriddedPerm]):
+    """
+    A class for a specific slidding strategy. The init probably gets the index of both
+    column you slidding.
+    """
```
christianbean, line 135: You can remove 'probably' now.

tilings/strategies/sliding.py
```diff
@@ -0,0 +1,276 @@
...
+            )
+        return (
+            self.maps.g_inv(
+                gp_slide_inverse(self.maps.g_map(gp), self.av_12, self.av_123)
+            ),
+        )
+
+    @classmethod
+    def from_dict(cls, d: dict) -> "SlidingStrategy":
+        raise NotImplementedError()
```
christianbean, line 202: We'll need to implement the json methods before accepting. I'll add some things to test files that you need to make sure pass.

tilings/strategies/sliding.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+class SlidingFactory(StrategyFactory[Tiling]):
+    """
+    A strategy factory is producing all the valid strategies of a given type that can
+    apply to the given tiling.
+
+    Here you want the call the method to return all the valid sliding strategy for the
+    given tiling.
+    """
```
christianbean, line 231: This need rewritten.

christianbean (changes requested):

christianbean (approved):

tilings/strategies/sliding.py
```diff
@@ -0,0 +1,286 @@
...
+    A class for a specific slidding strategy. The init gets the index of both
+    column you slidding.
+    """
+
+    def __init__(
+        self,
+        av_12_column: int,
+        av_123_column: int,
+        sliding: Sliding,
```
enadeau, line 135: Isn't that a bit too much that the strategy needs to store the whole sliding algorithm? If I compare with the fusion algorithm, only the column or row fused is stored and the algorithm can be recomputed if needed.

tilings/strategies/sliding.py
```diff
@@ -0,0 +1,286 @@
...
+                yield SlidingStrategy(*pair, sliding, _AdditionalMaps.rotate90(r))
+            tiling = tiling.reverse()
+            sliding = Sliding(tiling)
+            for pair in sliding.slidable_pairs():
+                yield SlidingStrategy(
+                    *pair, sliding, _AdditionalMaps.rotate90_and_reverse(r)
+                )
+
+    def __repr__(self) -> str:
+        return str(self)
```
enadeau, line 271: Can you implement a proper repr? That could come in handy for debugging

enadeau (changes requested):
I've left a few comments most are easy fix but there's a most substantial one about the argument for the sliding strategy. Perhaps we can discuss it in meeting

tilings/strategies/sliding.py
```diff
@@ -137,48 +130,52 @@ class SlidingStrategy(DisjointUnionStrategy[Tiling, GriddedPerm]):
...
-        self.av_123 = av_123_column
-        self.sliding = sliding
-        self.maps = maps
+        self.av_12 = av_12
+        self.av_123 = av_123
+        self.symmetry_type = symmetry_type
 
-    def decomposition_function(self, _tiling: Tiling) -> Tuple[Tiling]:
-        """Return the slidded tiling if it slides, otherwise None."""
+    def __call__(
```
enadeau, line 139: You don't actually need to define that, it's already defined on the super class

enadeau (changes requested):

enadeau (approved):

PR #288, opened 2020-11-26
dont initialise when removing/adding assumptions

christianbean:


enadeau (commented):
My only comment is I'm not sure `clean_assumptions` needs to be made public since it's only used internally

christianbean:
> My only comment is I'm not sure `clean_assumptions` needs to be made public since it's only used internally

just my way around the pylint error as I'm not using it on self in the method, but Tiling instance.

enadeau (approved):
Pylint is stupid sometime

PR #297, opened 2021-01-11
Fix assumption ordering

### LINKED ISSUE #296: Broken order on assumption
Looks like the ordering on assumption is broken sometime. For example the snippet below doesn't crash
```py
SumComponentAssumption((GriddedPerm((0,), ((1, 1),)), GriddedPerm((0,), ((2, 0),))))
SkewComponentAssumption((GriddedPerm((0,), ((1, 1),)), GriddedPerm((0,), ((2, 0),)), GriddedPerm((0,), ((2, 2),))))
assert ass1 < ass2
assert ass1 > ass2
```

enadeau:
Fixes #296

### DIFF
--- a/tests/test_assumptions.py
+++ b/tests/test_assumptions.py
@@ -1,5 +1,6 @@
 import json
 import os
+from operator import xor
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


christianbean (approved):
Good catch, it should have been an and, not an or.

enadeau:
> Good catch, it should have been an and, not an or.

The and is actually not the right thing to do because you still can't compare two assumption of a different type. For example comparing a skew and a sum assumption none will be the smallest because the first clause of the `and` can never be satisfied.

Anyway the tuple approach is working so merging it

PR #299, opened 2021-01-14
reverse fusion counting

enadeau:


tilings/strategies/fusion/constructor.py
```diff
@@ -545,3 +548,188 @@ def random_sample_sub_objects(
...
+            new_param = self.forward_map(param)
+            new_value = value - child_terms[self.a_map(param)]
+            assert new_param not in terms or new_value == terms[new_param]
+            terms[new_param] = new_value
+        return terms
+
+    def get_equation(
+        self, lhs_func: sympy.Function, rhs_funcs: Tuple[sympy.Function, ...]
+    ) -> sympy.Eq:
+        raise NotImplementedError
```
christianbean, line 717: Should we make this use the same method as the normal fusion constructor?

tilings/algorithms/fusion.py
```diff
@@ -3,9 +3,9 @@
 
 # TODO: this file needs to be type checked!
```
christianbean, line 4: Its done, hooray!

tilings/strategies/fusion/constructor.py
enadeau: Will make a fix on css to change the get equation of reverse rule to use the original rule

tilings/algorithms/fusion.py
enadeau: Fixed

enadeau:
I think it is now ready to go

christianbean (approved):

PR #301, opened 2021-01-25
guided search checks for empty tiling

enadeau:


tilings/tilescope.py
```diff
@@ -143,7 +143,11 @@ def __init__(
         self.tilings = frozenset(tilings)
         super().__init__(basis, pack, *args, **kwargs)
         for t in self.tilings:
-            self._add_to_queue(self.classdb.get_label(t))
+            class_label = self.classdb.get_label(t)
+            is_empty = t.is_empty()
+            self.classdb.set_empty(class_label, is_empty)
```
christianbean, line 148: If you are going to use the classdb you should use `self.classdb.is_empty`

christianbean (changes requested):

tilings/tilescope.py
enadeau: Done

PR #302, opened 2021-01-26
Invalid min point for fusion algo

enadeau:
If you call the min number of point on the fusion algorithm before checking if
it is fusable you get an incorrect answer. Just added the failing test for now
but will fix after merging reverse fusion


enadeau:
Finally took the time to fix that. Ready to merge

christianbean (approved):

PR #303, opened 2021-02-09
reverse fusion bug

enadeau:


christianbean (approved):
I see nothing obviously wrong, but hare to say if it is good as I've lost track of the goal of the function changed. I trust you.

PR #304, opened 2021-02-10
adds rearrange factory

jaypantone:


enadeau:
Should we also change the make tracked function to add the rearrange factory?

jaypantone:
Yes, great point, and also I broke the readme.

enadeau (approved):

PR #307, opened 2021-02-15
fix json encoding of rearrange strategy

### LINKED ISSUE #305: Broken `from_dict` in `RearrangeAssumptionStrategy`
The assert in this function is raised because `d` still contains `workable: True`. I'm not sure where that is supposed to be sent.
```py
    @classmethod
    def from_dict(cls, d: dict) -> "RearrangeAssumptionStrategy":
        assumption = TrackingAssumption.from_dict(d.pop("assumption"))
        sub_assumption = TrackingAssumption.from_dict(d.pop("sub_assumption"))
        assert not d
        return cls(assumption, sub_assumption)
```

enadeau:
Closes #305

### DIFF
--- a/tests/strategies/test_json_encoding.py
+++ b/tests/strategies/test_json_encoding.py
@@ -6,7 +6,7 @@ import pytest
 from comb_spec_searcher import Strategy
 from permuta import Perm
 from permuta.misc import DIR_EAST, DIR_NORTH, DIR_SOUTH, DIR_WEST, DIRS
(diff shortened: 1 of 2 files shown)


jaypantone (approved):
Thanks! So "workable" just doesn't need to be sent anywhere?

enadeau:
Yes should not be stored in the json in the first place since it's not needed to rebuild the strategy anyway

PR #309, opened 2021-03-17
Rule to sanity check

enadeau:
Rules that fails sanity check. The fix for the counting will be introduced in
the next version of css


enadeau:
Is now included in #314

PR #314, opened 2021-04-14
add a sanity check failure rule

christianbean:
Added to the test file. We need to investigate why this is broken.


enadeau:
The fix is in this pull request: PermutaTriangle/comb_spec_searcher/pull/268

Should be able to merge with the next release of css

PR #315, opened 2021-04-16
Fix isolated fusion and Guided Searcher

jaypantone:


christianbean (approved):

enadeau:
Could you add the tiling of the bug to the fusion tests?

jaypantone:
> Could you add the tiling of the bug to the fusion tests?

Maybe... Will have to hunt it down.

jaypantone:
Added the test case.

enadeau (approved):

PR #320, opened 2021-04-22
Old strategy JSON

enadeau:
This introduces a new series of test where we make sure we don't brake backward
compatibility with JSON.

When we run into a backward compatibility problem we can add the JSON to the new
file to get it tested.


christianbean (approved):
This is a good addition

PR #333, opened 2021-05-17
empty tiling is not an atom

christianbean:


enadeau (approved):
With a changelog update and a test! It really checks all the boxes!

PR #336, opened 2021-05-19
shifts locally factorable

enadeau:
I've changed the locally factorable shift computation to rely on the expansion instead of the subobstruction inferral. To speed up the computation of the shift I've created a verification strategy that verifies any tiling that doesn't have a cell that is the root. This will reduce the amount of expansion needed to compute the shifts and the children.


christianbean:
I disagree with the solution. I think you should make a StrategyFactory, which Factors if possible, inserts a pattern if necessary, and only once all of the cells are not the root class returns a verification rule. These rules should all have ignore_parent=True and workable=False.

Note, you can still put a StrategyFactory in the verification strategies.

enadeau:
I don't think this is the way to move forward. It's especially problematic for the recomputing of the strategies since you'll then have arbitrary strategies in the searcher. There's no rush to merge this, so we can discuss that next meeting.

PR #355, opened 2021-09-23
new assumption cell map

enadeau:
- make is_identity param optional in map
- add compose function for map
- add __eq__ to map
- make map hashable
- make the tracking assumption use the cell map
- move map outside of algo
- adding test for non-crossing map
- improve map string


enadeau:
Approved by Christian

PR #357, opened 2021-09-23
add tilings contain method

enadeau:


tilings/tiling.py
```diff
@@ -1838,6 +1838,15 @@ def __ne__(self, other: object) -> bool:
             or self.assumptions != other.assumptions
         )
 
+    def __contains__(self, gp: GriddedPerm) -> bool:
```
christianbean, line 1841: This would be better more like:
gp.avoids(*obs) and all(gp.contains(*req) for req in reqs)

enadeau: Absolutely, this is much cleaner.

christianbean (approved):

PR #359, opened 2021-09-29
Change OneByOneVerificaiton so that is only verifies subclasses

### LINKED ISSUE #354: One by one verification should only verify subclass
Make one by one only verify subclasses of the given basis and not any class that is not exactly the given one.

enadeau:
Closes #354


christianbean (approved):
Looks good! Can you fix the change log, then merge.

PR #360, opened 2021-10-05
Row columns separation strategy

enadeau:
- typing rcs algo
- introducing the cell map class
- use the CellMap class in row col sep algorithm
- add map_tiling method
- allow crossing map for preimage counter
- map preimage counter
- make param counter iterable
- implement map parameters
- map inverse method and to_row_col_map
- simplify projection computation
- move param mapping stuff to the map class
- fix the strategy


christianbean (approved):
Read through with Emile, I'm happy.

PR #377, opened 2021-11-01
Remove unnecessary method redefintion

enadeau:
- remove duplicated method


tilings/map.py
```diff
@@ -207,9 +208,10 @@ def __init__(
         self._row_map = row_map
         self._col_map = col_map
         self._is_identity = is_identity
+        print("BBB")
```
christianbean, line 211: you want to remove these prints

enadeau: Ah, that is embarrassing. Fixed it

PR #380, opened 2021-11-09
display

enadeau:
Create a class that take care of displaying the tiling both for the ascii and
the html version.

Changed the labelling to using letter and being consistent across tiling over a
session.

- make the row_map and col_map public via property
- use displayer class to build tiling string
- special symbol for empty tiling
- move html code to display algorithm


enadeau:
The bug is caused when trying to display this tiling:
```py
    t = Tiling(
        obstructions=(),
        requirements=(),
        parameters=(
            ParameterCounter(
                (
                    PreimageCounter(
                        Tiling(obstructions=(), requirements=(), parameters=()),
                        RowColMap({}, {}),
                    ),
                )
            ),
        ),
    )
```

It fails to compute the extra obs and reqs for the preimage.

christianbean (approved):

PR #398, opened 2021-11-23
ensure that we remove duplicate requirements

christianbean:
Moreover, if a requirement list R1 implies another requirement list R2, we should only keep the list R1. This does not seem to happen on develop at the moment. Pushed a fix with a test.


tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -124,6 +125,28 @@ def minimal_reqs(
                 return (tuple(),)
         return tuple(minimized_requirements)
 
+    def remove_redundant(self, requirements: List[Requirement]) -> List[Requirement]:
+        relevant_reqs: List[Requirement] = []
+        for idx, requirement in enumerate(requirements):
```
jaypantone, line 130: "requirement" here is a requirement list?

tilings/algorithms/gridded_perm_reduction.py
```diff
@@ -124,6 +125,28 @@ def minimal_reqs(
...
+            if self._requirement_implied_by_some_requirement(
+                requirement,
+                chain(
+                    islice(requirements, idx),
+                    [
+                        gps
+                        for gps in islice(requirements, idx + 1, None)
+                        if gps != requirement
+                    ],
+                ),
```
jaypantone, line 142: Is the use of islice and chain to keep this a generator? If so, should the second argument use () instead of [] so it too is a generator?

jaypantone: Also why do we need to check gps != requirement? Equality would mean two identical requirements lists, and that should also be detected as redundant, right?

tilings/algorithms/gridded_perm_reduction.py
christianbean: yes, its type is precisely is `Tuple[GriddedPerm, ...]`

tilings/algorithms/gridded_perm_reduction.py
christianbean: it should be (), good catch.

the purpose of the `!=` is to ensure that if two are exactly the same, we leave one of them - the setup ensures we always leave the first one.

jaypantone: Ah, very smart. Nice.

jaypantone (approved):

PR #400, opened 2021-11-25
a new approach to tracked searcher!

christianbean:
We now create a new `TrackedQueue`. The idea is this is a list of queues, and when a label is added it is added to the same queue that the underlying label was added to.


jaypantone:
Would be good to have a `levels_completed` function like the default queue does.

christianbean:
> Would be good to have a `levels_completed` function like the default queue does.

I can add this. I will also try to add more stats to the status update to make it more intuitive about what is going on.

PR #404, opened 2021-12-02
the cell insertion pack

christianbean:


jaypantone:
Should this pack get the new remove_requirement strategy? Any value there?

christianbean:
> Should this pack get the new remove_requirement strategy? Any value there?

Since any 1x1 whose basis is not the root we verify, and we cell insert in all ways into the root, it doesn't add anything.

jaypantone (approved):

PR #405, opened 2021-12-02
add the remove req factory to the locally factorable pack

christianbean:
Relies on [PR #316 on CSS](https://github.com/PermutaTriangle/comb_spec_searcher/pull/316)


jaypantone (approved):
I see this is just used in the locally factorable expansion pack. Probably good to not try to use it as a general strategy because it will make so many rules?

christianbean:
> I see this is just used in the locally factorable expansion pack. Probably good to not try to use it as a general strategy because it will make so many rules?

We use a version of this with parameters, and yes it does make so many rules!

jaypantone (approved):

PR #431, opened 2022-03-11
Trackedclassdb

christianbean:


jaypantone:
Also contains the changes from PR for pointing, by accident.

jaypantone:
I've intentionally left in a few commented out lines of code so you can more easily compare the old code with the new code (not sure the diff will do it well), and then I can remove them right before we merge.

tilings/tilescope.py
```diff
@@ -78,6 +87,10 @@ def __init__(
             expand_verified=expand_verified,
             debug=debug,
         )
+        if isinstance(classdb, TrackedClassDB):
```
enadeau, line 90: Make this a flag we can change on css

PR #432, opened 2022-03-14
Improving initialiser

christianbean:
Do the potentially large product in a smarter way.  

Also includes the change to the pattern placement pack.


jaypantone:
I might need the `_product_requirements` explained to me...

PR #452, opened 2022-05-03
Insane factoring

enadeau:


jaypantone:
This gave a "spec" for 1324 that we couldn't do anything with, so it's probably way too powerful.

PR #453, opened 2022-05-04
pointing strategy, unfusion strategy, disjoint fusion strategy, a dummy strategy, and allowing directionless pp

christianbean:


tilings/algorithms/requirement_placement.py
```diff
@@ -362,6 +362,9 @@ def place_point_of_req(
         for cell in sorted(cells):
             stretched = self._stretched_obstructions_requirements_and_assumptions(cell)
             (obs, reqs, ass) = stretched
+            if direction == DIR_NONE:
+                res.append(self._tiling.__class__(obs, reqs, ass))
+                continue
```
ulfarsson, line 367: This is lovely

tilings/strategies/fusion/disjoint_fusion.py
```diff
@@ -0,0 +1,95 @@
...
+        )
+
+    def fusion_algorithm(self, tiling: Tiling) -> Fusion:
+        return Fusion(tiling, row_idx=self.row_idx, col_idx=self.col_idx, tracked=False)
+
+    def decomposition_function(self, comb_class: Tiling) -> Tuple[Tiling, ...]:
+        algo = self.fusion_algorithm(comb_class)
+        if algo.fusable():
+            fused_tiling = algo.fused_tiling()
+            point_count = sum(algo.min_left_right_points())
```
ulfarsson, line 25: This point counting should be discussed later (made a not in the agenda). Can be merged now and we worry about generalizing to more positive tilings later

tilings/strategies/fusion/disjoint_fusion.py
```diff
@@ -0,0 +1,95 @@
...
+        self.col_idx = col_idx
+        self.row_idx = row_idx
+        super().__init__(
+            ignore_parent=False, inferrable=True, possibly_empty=False, workable=True
+        )
+
+    def fusion_algorithm(self, tiling: Tiling) -> Fusion:
+        return Fusion(tiling, row_idx=self.row_idx, col_idx=self.col_idx, tracked=False)
+
+    def decomposition_function(self, comb_class: Tiling) -> Tuple[Tiling, ...]:
```
ulfarsson, line 21: Note that we originally drew this strategy as A -> (B, C) where A is a 1x2 domino. C is the unfusion of the first column. B is the pointing in the first cell. The strategy here is implemented C -> (A, B). Note also that the strategy can handle larger tilings, in which case you need to point  to multiple cell. See https://discord.com/channels/383038476553158657/383041809980325896/971080895312326736

tilings/strategies/pointing.py
```diff
@@ -0,0 +1,123 @@
...
+        comb_class: Tiling,
+        children: Optional[Tuple[Tiling, ...]],
+    ) -> Tuple[int, ...]:
+        if children is None:
+            children = self.decomposition_function(comb_class)
+        assert children is not None
+        return tuple(0 for _ in children)
+
+    def decomposition_function(self, comb_class: Tiling) -> Tuple[Tiling, ...]:
+        cells = comb_class.active_cells - comb_class.point_cells
```
ulfarsson, line 37: This is slightly wrong, should only subtract point cells which are in their own row and column. Christian will fix this before merging

tilings/strategies/unfusion.py
```diff
@@ -0,0 +1,317 @@
+from collections import Counter
```
ulfarsson, line 1: Same thing here. We unfuse either every row, or every column to make sure that no extra assumptions are needed. We can discuss later whether to unfuse a single row, or single column, but then assumptions are needed.

This strategy will be applied to every tiling so make sure to use it as an expansion strategy. Will discuss later (agenda) whether to give it an option to be only applied to "small" tilings

ulfarsson (approved):
Approved. See comments. I think we can merge this and discuss some extras next meeting.

Christian needs to add a quick hotfix for point cells. Otherwise Jay can merge

PR #456, opened 2022-05-06
reversible if parent has all the assumptions for interleaving

christianbean:


tilings/strategies/factor.py
```diff
@@ -244,6 +244,12 @@ def __init__(
         self.tracked = tracked
         self.cols, self.rows = self.interleaving_rows_and_cols(self.partition)
 
+    def is_two_way(self, comb_class: Tiling) -> bool:  # type: ignore
+        return self.is_reversible(comb_class)
+
+    def is_reversible(self, comb_class: Tiling) -> bool:  # type: ignore
+        return not bool(self.assumptions_to_add(comb_class))
```
ulfarsson, line 251: Checking whether there are any assumptions to be added. If none have to be added they are all there and you can reverse the rule

PR #457, opened 2022-05-06
make sure to check both cells are monotone

christianbean:


ulfarsson (approved):
Fixing checking the same cell twice to checking that both cells are monotone

Issue #454, opened 2022-05-05
NotImplementedError: removing exact components


jaypantone:
To recreate in about 90 seconds, switch to `insert-factorable-reqs` branch on `Tilings` and run
```py
from tilings.tilescope import TileScopePack, TrackedSearcher
from comb_spec_searcher.rule_db import RuleDBForest

basis = "0213"
pack = TileScopePack.all_the_strategies(3).make_fusion(tracked=False).make_fusion(component=True, tracked=False).kitchen_sinkify(3, 0, False).make_tracked()
css = TrackedSearcher(basis, pack, ruledb=RuleDBForest(), expand_verified=True, max_assumptions=5)
css.auto_search(status_update=10)
```

Traceback:
```py
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
<ipython-input-9-61113f09af86> in <module>
      5 pack = TileScopePack.all_the_strategies(3).make_fusion(tracked=False).make_fusion(component=True, tracked=False).kitchen_sinkify(3, 0, False).make_tracked()
      6 css = TrackedSearcher(basis, pack, ruledb=RuleDBForest(), expand_verified=True, max_assumptions=5)
----> 7 css.auto_search(status_update=10)

~/Dropbox/Research/Active/2017-44-ATRAP/repos/comb_spec_searcher/comb_spec_searcher/comb_spec_searcher.py in auto_search(self, **kwargs)
    494         start_string += self.run_information()
    495         logger.info(start_string)
--> 496         spec_rules = self._auto_search_rules(**kwargs)
    497         if spec_rules is not None:
    498             specification = CombinatorialSpecification(self.start_class, spec_rules)

~/Dropbox/Research/Active/2017-44-ATRAP/repos/comb_spec_searcher/comb_spec_searcher/comb_spec_searcher.py in _auto_search_rules(self, max_expansion_time, perc, smallest, status_update)
    527         while expanding:
    528             expanding, status_start = self._expand_classes_for(
--> 529                 expansion_time, status_update, status_start, auto_search_start
    530             )
    531             spec_search_start = time.time()

~/Dropbox/Research/Active/2017-44-ATRAP/repos/comb_spec_searcher/comb_spec_searcher/comb_spec_searcher.py in _expand_classes_for(self, expansion_time, status_update, status_start, auto_search_start)
    576                 last_label = label
    577             if self.expand_verified or not self.ruledb.is_verified(label):
[...]

PR #460, opened 2022-05-09
count component behave like disjoint in shift algo

christianbean:


christianbean:
fixes issue #454

enadeau (approved):
Makes perfect sense. Good to go

PR #472, opened 2022-05-31
Obs inf early yield

jaypantone:


tilings/algorithms/gridded_perm_generation.py
```diff
@@ -32,17 +32,20 @@ class GriddedPermsOnTiling:
     built by inserting points into the minimal gridded permutations.
     """
 
-    def __init__(self, tiling: "Tiling"):
+    def __init__(self, tiling: "Tiling", yield_non_minimal=False):
```
christianbean, line 35: this should be typed:
`yield_non_minimal: bool = False`
in `new_obs` method too.

tilings/strategies/obstruction_inferral.py
```diff
@@ -117,7 +117,9 @@ def new_obs(self, tiling: Tiling) -> Sequence[GriddedPerm]:
         """
         Returns the list of new obstructions that can be added to the tiling.
         """
-        return AllObstructionInferral(tiling, self.maxlen).new_obs()
+        return AllObstructionInferral(tiling, self.maxlen).new_obs(
+            yield_non_minimal=False
+        )
```
christianbean, line 122: The deafult should be True

PR #481, opened 2022-06-15
adds default option

jaypantone:


jaypantone:
Without a default option for the new allow_factorable_insertions flag, old specs won't load.

christianbean (approved):

PR #483, opened 2022-06-15
one by one looks up permpal

christianbean:


christianbean:
Relies on https://github.com/PermutaTriangle/comb_spec_searcher/pull/331

PR #487, opened 2022-06-28
stop expanding av(012), and return invalid op for comp ver

christianbean:


jaypantone:
This is annoyingly going to require changing some of the test cases.

PR #494, opened 2022-07-27
ensure obs are minimised in point placements

christianbean:


tilings/algorithms/requirement_placement.py
```diff
@@ -377,15 +377,19 @@ def place_point_of_req(
...
                     res.append(self._tiling.__class__(obs + rem_req, reqs, ass))
                 continue
-
             forced_obs = self.forced_obstructions_from_requirement(
                 gps, indices, cell, direction
             )
+            forced_obs = [
+                o1
+                for o1 in forced_obs
+                if not any(o2 in o1 for o2 in filterfalse(o1.__eq__, forced_obs))
```
jaypantone, line 386: Why is this different / better than
`if not any(o2 in o1 for o2 in forced_obs if o2 != o1)`
?

tilings/algorithms/requirement_placement.py
```diff
@@ -377,15 +377,19 @@ def place_point_of_req(
...
+            ]
             reduced_obs = [o1 for o1 in obs if not any(o2 in o1 for o2 in forced_obs)]
-
+            reduced_obs.extend(filterfalse(reduced_obs.__contains__, forced_obs))
```
jaypantone, line 389: So we have [minimal forced obs] + [things in obs that don't contain a forced ob]. Is there a chance that something in "obs" *is contained in* something in "forced obs"? If so, this isn't accounted for.

Also, is this line equivalent to
`reduced_obs.extend(fob for fob in forced_obs if fob not in reduced obs)`
? Would it be better to just extend it with all of forced_obs, then cast it to a set to remove duplicates?

christianbean: forced obs come from a requirement, so this is why we do it here, rather than just in the tiling initialiser.

christianbean: In my test, using a set was slower

jaypantone (approved):

PR #503, opened 2023-01-20
check in decomp func of sliding

christianbean:
Previous broken rules will now raise `StrategyDoesNotApply` when calling the decomposition function.


jaypantone (approved):
changelog?

PR #513, opened 2023-02-23
a strategy for forcing assumptions to be empty

christianbean:


jaypantone:
Testing on a few things. Haven't found anything too exciting yet. If we do merge this, we should add it to the kitchen sink.

jaypantone:
I don't have good examples of this strategy being useful, but I still think we should merge it.

